PR #482, opened 2026-06-05
Upload your own validation plan (secure import endpoint + editor button)

### LINKED ISSUE #481: Feature: upload your own validation plan (with strict server-side sanitization)
## Summary

Add an **Upload** control to the workflow editor's **Validation Plan** section so a user can import their own validation plan from a JSON file, instead of only being able to LLM-**Regenerate** it or hand-edit checks one at a time.

## Motivation

- The current plan is either LLM-generated (non-deterministic, only sees the first ~2k chars of each prompt) or hand-built check-by-check in the UI.
- Teams want **version-controlled, reviewable, reusable** validation plans they can author offline (e.g. in a repo) and apply consistently across workflows — the AI4RA prompt-library ships exactly these per-workflow plans.
- An import path makes plans portable and reproducible.

## Scope

- **Frontend**: an "Upload" button next to "Regenerate" (and in the empty state). It opens a file picker, parses the JSON client-side, and POSTs the checks.
- **Backend**: a new `POST /workflows/{id}/validation-plan/import` endpoint that **strictly validates and sanitizes** the uploaded plan before persisting it.
- Accepts either a bare array of checks or a `{ "checks": [...] }` wrapper. Canonical check shape: `{name, description, category}` (category ∈ completeness | accuracy | content | formatting).

## Security requirements (must-have)

The uploaded file is **untrusted input**, so the import must not become an injection/abuse vector. The endpoint must:

- be **server-authoritative** — never trust client-side validation alone;
- require **manage rights** on the workflow and be **rate-limited**;
- **cap** the number of checks and the length of every string (bound stored size and the text that later flows into the LLM judge prompt at Validate time);
- restrict `category` to the known enum;
- **drop unknown keys** so nothing extra is ever stored or evaluated;
- **synthesize check ids server-side** (ignore client-supplied ids) to prevent id collision/forgery;
- **strip control characters** (defuse terminal/log-injection and prompt-injection framing);
- ensure rendered check text is treated as **text, not HTML** (no stored XSS);
- **not write any file to disk** — parse JSON in memory (no path-traversal / multipart-file surface).

Residual, documented: a check `description` is, by design, fed into the LLM judge prompt at Validate time (true of any plan, generated or uploaded). Length caps + control-char stripping reduce framing; the judge only emits PASS/FAIL/detail and executes nothing.

A PR implementing this (endpoint + sanitizer + frontend Upload button + tests) follows.

StringTheoryDev:
## What

Adds an **Upload** button to the workflow editor's **Validation Plan** section so users can import their own validation plan from a JSON file, with **strict server-side validation/sanitization**.

Closes #481.

## Why

Today a plan is either LLM-**Regenerated** (non-deterministic; only sees ~2k chars of each prompt) or hand-edited check-by-check. Teams want version-controlled, reviewable, reusable plans authored offline (e.g. in the AI4RA prompt-library) and applied consistently. See #481.

## Backend — treats the upload as untrusted input

- **New endpoint** `POST /workflows/{id}/validation-plan/import` — requires **manage** rights, **rate-limited** `10/min`.
- **`_sanitize_uploaded_checks()`** (`workflow_service.py`):
  - requires a **non-empty JSON array** within a hard **count cap (50)**;
  - keeps **only** `{id, name, description, category}` — **unknown keys are dropped** (never stored or evaluated);
  - **synthesizes `id` server-side** and **ignores any client id** (no id collision/forgery);
  - restricts `category` to the enum `{completeness, accuracy, content, formatting}` (with a small alias map, e.g. `format`→`formatting`);
  - **strips C0/DEL control characters** (defuses terminal/log-injection and prompt-injection framing) and **caps** `name` (120) / `description` (2000);
  - raises friendly **per-check** `ValueError`s → `400` (e.g. "Check #3: invalid category 'foo'").
- `import_validation_plan()` persists the sanitized plan.
- `ImportValidationPlanRequest` accepts raw objects so the service can return friendly errors instead of opaque 422s.
- **No file is written to disk** — JSON is parsed in memory (no path-traversal / multipart-file surface). Output is rendered as **text** in the UI (no stored XSS; no `dangerouslySetInnerHTML`).

## Frontend

- `importValidationPlan()` API client.
- **Upload** button beside **Regenerate** and in the empty state; a hidden file input; a client-side **256 KB** size guard + `JSON.parse`; accepts a bare array or `{ "checks": [...] }`; success/error toasts. The server remains authoritative.

## Tests

`TestSanitizeUploadedChecks` (9 cases): valid passthrough, id synthesis, **client-id ignored**, **unknown-key drop**, category alias + invalid rejection, **control-char stripping**, non-list/empty/oversize-count rejection, missing/non-string-field rejection, oversize-string rejection.

```
cd backend && uv run pytest tests/test_workflow_service.py -q -k SanitizeUploadedChecks
```

## Security summary

Server-authoritative validation; auth + rate limit; count/length caps; enum-restricted category; unknown keys dropped; server-synthesized ids; control-char stripping; text-only rendering; no disk writes. **Residual (documented):** a check `description` is, by design, fed into the LLM judge prompt at Validate time (true of any plan); caps + stripping reduce framing, and the judge only emits PASS/FAIL/detail and executes nothing.

> Note: CI runs the typecheck/test suites. The backend changes were `py_compile`-checked and the sanitizer logic unit-verified locally; the frontend mirrors existing editor patterns (couldn't run `tsc`/`pytest` in the authoring env — no `uv`/`node_modules`).

### DIFF
--- a/backend/app/routers/workflows.py
+++ b/backend/app/routers/workflows.py
@@ -24,6 +24,7 @@ from app.schemas.workflows import (
     BatchStatusResponse,
     CreateTempDocumentsRequest,
     CreateWorkflowRequest,
+    ImportValidationPlanRequest,
     ReorderStepsRequest,
     RunWorkflowRequest,
     TestStepRequest,
@@ -1383,6 +1384,24 @@ async def update_validation_plan(
 async def generate_validation_plan(request: Request, workflow_id: str, user: User = Depends(get_current_user)):
     try:
         checks = await svc.generate_validation_plan(workflow_id, user=user)
+        return ValidationPlanResponse(checks=checks)
+    except ValueError as e:
+        raise HTTPException(status_code=400, detail=str(e))
+
+
+@router.post("/{workflow_id}/validation-plan/import", response_model=ValidationPlanResponse)
+@limiter.limit("10/minute")
+async def import_validation_plan(
+    request: Request,
+    workflow_id: str,
+    req: ImportValidationPlanRequest,
+    user: User = Depends(get_current_user),
+):
+    """Import a user-uploaded validation plan. The payload is strictly
+    validated and sanitized server-side (see
+    workflow_service._sanitize_uploaded_checks) before it is persisted."""
+    try:
+        checks = await svc.import_validation_plan(workflow_id, req.checks, user=user)
         return ValidationPlanResponse(checks=checks)
     except ValueError as e:
         raise HTTPException(status_code=400, detail=str(e))
--- a/backend/app/schemas/workflows.py
+++ b/backend/app/schemas/workflows.py
@@ -137,6 +137,14 @@ class UpdateValidationPlanRequest(BaseModel):
     checks: list[ValidationCheckDefinition]
 
 
+class ImportValidationPlanRequest(BaseModel):
+    # Raw, user-uploaded checks. Accepted as arbitrary objects so the service
+    # can strictly validate/sanitize them and return friendly per-check errors
+    # instead of opaque 422s. Never trusted as-is — see
+    # workflow_service._sanitize_uploaded_checks.
+    checks: list[Any]
+
+
 class ValidationPlanResponse(BaseModel):
     checks: list[ValidationCheckDefinition]
 
--- a/backend/app/services/workflow_service.py
+++ b/backend/app/services/workflow_service.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import datetime
 import logging
+import re
 import uuid as uuid_mod
 from typing import TYPE_CHECKING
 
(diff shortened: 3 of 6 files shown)


ViaJables:
Closing for now — this is a direction call, not a quality issue with the PR.

The Validation Plan editor this hangs off still exists, and the sanitizer targets the current canonical `{id, name, description, category}` schema, so the change still applies cleanly. But after the validation-uplift rebuild the **check-plan validation is a diagnostic surface** — the authoritative scorer is now the auto-tune flow. We don't want to keep extending the check-plan editor (an extra import endpoint + a sanitizer to maintain) on a surface we're de-emphasizing.

Happy to revisit if version-controlled, offline-authored plans become a concrete need. The security model here (in-memory parse, server-authoritative sanitization, control-char stripping, id synthesis) was nicely done and worth pulling back in if we do.

PR #484, opened 2026-06-06
Download the full validation report (Markdown / JSON)

StringTheoryDev:
## What

Adds a way to **download the validation report** (Markdown or JSON) from the editor's Validation Grade panel. Today the grade, per-check PASS/FAIL, and evidence are only visible in-app — there's no exportable artifact.

Closes #483.

## Why

An RA who runs Validate and gets "Grade D, 2 failures" needs to save/share *where* it failed (attach to an award record, send to a PI, diff over time). The full result is already persisted in `ValidationRun.result_snapshot`; this just exposes an export path. See #483.

## Backend

- **New endpoint** `GET /workflows/{id}/validation-report?format=md|json` — fetches the latest persisted `ValidationRun` and returns it as a downloadable file (`Content-Disposition: attachment`). 404 when there are no runs yet; 400 on a bad `format`.
- **`_format_validation_report()`** — a **pure, unit-testable** renderer:
  - **Markdown** (default): title, grade + score, summary, runs/stability, then each check as `### [STATUS] Name — _category_` with its evidence/detail, plus a per-run line when a check was inconsistent across runs.
  - **JSON**: structured report (grade, score, per-check status/detail/category, stability) for programmatic use.
- **`build_validation_report()`** fetches (auth via `get_authorized_workflow`) + renders.
- **`quality_service.get_latest_validation_run()`** — public accessor for the latest run.

## Frontend

- `validationReportUrl()` API helper.
- **Report** (Markdown) + **JSON** buttons in the Validation Grade panel, shown once a grade exists. Cookie-auth GET via `window.open`, mirroring the existing `exportWorkflowUrl` / download pattern.

## Tests

`TestFormatValidationReport` (4 cases): markdown rendering (grade, per-check status, evidence, category, per-run line), JSON rendering, empty-checks, and filename slugify.

```
cd backend && uv run pytest tests/test_workflow_service.py -q -k FormatValidationReport
```

## Notes

- Independent of #480 / #482 (branched from `main`).
- Backend `py_compile`-checked and the pure renderer unit-verified locally; frontend mirrors existing patterns (couldn't run `tsc`/`pytest` in the authoring env — CI will).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #480, opened 2026-06-05
Make validation-plan evaluation schema-tolerant (fixes Validate KeyError)

StringTheoryDev:
## What

Makes validation-plan evaluation **schema-tolerant** so a workflow whose `validation_plan` uses the alternate `{check_id, check_name, check_description, check_type}` export schema no longer crashes **Validate** with `KeyError`.

Fixes #479.

## Why

The runtime/editor canonical check schema is `{id, name, description, category}`, but `export_import_service` copies an imported plan verbatim. When that plan uses the `{check_id, …}` shape, `_evaluate_checks_against_output` (and the multi-run merge / no-results path) index `c["id"]`/`c["name"]` and raise `KeyError: 'id'`. See #479 for the full trace.

## Changes

- **`backend/app/services/workflow_service.py`**
  - Add `_normalize_validation_plan(plan)` — coerces either schema into the canonical `{id, name, description, category}` shape:
    - `id` ← `id` | `check_id` (stable across calls; only synthesized via `uuid4` when neither is present)
    - `name` ← `name` | `check_name` (capped at 60 chars)
    - `description` ← `description` | `check_description`
    - `category` ← `category` | `check_type` mapped through `_VALIDATION_CATEGORY_ALIASES` (`format`→`formatting`; `arithmetic`/`consistency`/`correctness`/`constraints`/`hallucination`→`accuracy`; `presence`→`completeness`; unknown→`content`)
    - non-dict / malformed entries are skipped, never raising
  - Apply it in `validate_workflow` (the crash site) and `get_validation_plan` (so imported/alternate plans display and edit correctly).
- **`backend/tests/test_workflow_service.py`** — `TestNormalizeValidationPlan` with 6 cases: canonical passthrough, alternate-schema mapping, `check_type` alias table, id synthesis when absent, malformed-entry skipping, and `None`/empty input.

## Safety / scope

- Pure normalization on **read** — nothing is persisted or mutated, and well-formed canonical plans pass through byte-identical.
- No API contract change: `get_validation_plan` still returns `{id, name, description, category}` objects (now guaranteed even for imported plans).

## Test

```
cd backend && uv run pytest tests/test_workflow_service.py -q -k NormalizeValidationPlan
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #487, opened 2026-06-06
Ground the validation judge in real document text, not UUIDs

### LINKED ISSUE #486: Validation judge grades against document UUIDs, not document text (accuracy/completeness checks are blind)
## Summary

The workflow-validation **judge evaluates against document UUIDs instead of the document text**, so accuracy/grounding and completeness checks are unreliable for every workflow whose input arrives via `workflow_documents` (i.e. all document-driven workflows). The judge literally reports the source as "hash strings."

## Root cause (confirmed against live run data)

`_evaluate_checks_against_output` (`workflow_service.py`) builds its "Source Document Text (ground truth)" by scanning the run's `steps_output` for a `Document`/`AddDocument` step (lines ~1466–1493). But the `Document` trigger step's stored `output` is a **list of document UUIDs, not text**:

```
steps_output["Document"].output == ["EE11829E…","EFF22544…","79D5FE41…","15218AAF…"]
```

The builder hits the `isinstance(raw, list)` branch and does `"\n---\n".join(str(item) …)`, feeding the judge the **UUID hex strings** as "ground truth." The judge then can't find any extracted value in that "source" and declares everything ungrounded / hallucinated — e.g. *"$263,413 … do not appear in the provided source document (which only contains hash strings)."*

`retrieved_sources` is empty for these runs, and the real text lives in `SmartDocument.raw_text` keyed by `input_context.doc_uuids` (the same text the workflow itself ran on — `workflow_tasks.py:304` resolves it the same way).

## Impact

- **Accuracy** and **completeness** checks — the two highest-weighted dimensions — are graded **blind** for all document-driven workflows. A correct single-document extraction is falsely failed (its values "aren't in the source," because the judge's "source" is UUIDs). Grades on these workflows are not trustworthy.

## Fix

In `validate_workflow`, resolve each run's `input_context.doc_uuids` → `SmartDocument.raw_text` (plus any KB `retrieved_sources` chunks) and pass that to the judge as the ground-truth source, instead of the UUID-stringifying steps_output scan. The legacy steps_output path is kept as a fallback for runs that genuinely carry document text in a step.

Verified against live data: the 4 UUIDs from a real run resolve to 75,862 / 22,363 / 6,994 / 5,870 chars of real text.

A PR implementing this (resolver + judge override + unit tests) follows.

StringTheoryDev:
## What

Makes the workflow-validation judge evaluate against the **real document text**, not the document **UUIDs**. Fixes accuracy/completeness checks being graded blind for document-driven workflows.

Closes #486.

## Why

`_evaluate_checks_against_output` built its "Source Document Text (ground truth)" by scanning `steps_output` for a `Document`/`AddDocument` step. That step's `output` is a **list of document UUIDs**, so the judge was handed UUID hex strings as the "source" and reported it as *"hash strings"* — then failed every accuracy/grounding and completeness check because the extracted values "don't appear in the source." The two highest-weighted dimensions were effectively noise for all document-driven workflows. See #486 for the live trace.

## Changes (`backend/app/services/workflow_service.py`)

- **`_resolve_run_source_text(result)`** — resolves the run's `input_context.doc_uuids` to `SmartDocument.raw_text` (the same text the workflow ran on; same resolver the engine uses at `workflow_tasks.py:304`) and appends any KB `retrieved_sources` chunk previews. Returns `""` when nothing is recoverable.
- **`_evaluate_checks_against_output(..., source_text_override=None)`** — when an override is provided, it's used as the ground-truth source (capped 15k) instead of the UUID-stringifying `steps_output` scan. The legacy scan remains as a fallback for runs that genuinely carry document text in a step.
- **`validate_workflow`** — resolves per-run source text and passes it via `source_text_override`.

## Verification

Confirmed against a real run's data: its 4 document UUIDs resolve to **75,862 / 22,363 / 6,994 / 5,870** chars of real text the judge previously never saw (it only saw the 4 UUIDs).

## Tests

`TestResolveRunSourceText` (4 cases): UUID→`raw_text` resolution, skipping docs without `raw_text`, including KB `retrieved_sources`, and empty-when-no-sources.

```
cd backend && uv run pytest tests/test_workflow_service.py -q -k ResolveRunSourceText
```

## Notes

- Independent of #480 / #482 / #484 (branched from `main`).
- 15k source cap matches existing behavior; for very long single documents an early value past 15k could still be truncated — a smarter (section-targeted) extraction is a reasonable follow-up.
- This makes the judge reliable; it does not change what a workflow extracts. (E.g., feeding several *different* awards to one run still mixes them — that's a separate test-setup / batch-validation concern.)

### DIFF
--- a/backend/app/services/workflow_service.py
+++ b/backend/app/services/workflow_service.py
@@ -1395,7 +1395,10 @@ async def validate_workflow(workflow_id: str, user: User | None = None) -> dict:
                 for c in plan
             ]
         else:
-            run_checks = await _evaluate_checks_against_output(plan, output_text, wr.steps_output, wf_data)
+            src = await _resolve_run_source_text(wr)
+            run_checks = await _evaluate_checks_against_output(
+                plan, output_text, wr.steps_output, wf_data, source_text_override=src,
+            )
         all_run_checks.append(run_checks)
 
     # Merge multi-run results with consistency tracking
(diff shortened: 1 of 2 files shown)


ViaJables:
Carried forward — closing this PR branch (it targets pre-uplift `main` and doesn't cherry-pick onto the rewritten validation code).

Your fix has been **re-applied** on `feature/validation-uplift` in commit 4e1bc3c5: `_resolve_run_source_text()` resolves `input_context.doc_uuids` → `SmartDocument.raw_text` (+ KB `retrieved_sources`) and threads it into `_evaluate_checks_against_output` via a new `source_text_override` arg, with the legacy `steps_output` scan kept as fallback. Your `TestResolveRunSourceText` cases came along too (4/4 green). It'll land on `main` when the validation-uplift branch merges.

Confirmed the bug was still live on the new code before re-applying — the Document step output is still a list of UUIDs. Nice catch on the 75k-chars-the-judge-never-saw trace.

Issue #488, opened 2026-06-06
Batch validation: grade a workflow across a document set (per-document results), not a rolling last-3 window


StringTheoryDev:
## Summary

Vandalizer can **run** a workflow across many documents (batch mode), and it can **validate** a workflow's output — but there is **no batch *validation***: no way to grade a workflow across a *set* of representative documents and see a **per-document** pass/fail breakdown. Worse, the existing single-run validator, when fed batch runs, actively produces misleading grades. This issue proposes a first-class batch-validation capability and explains why it's needed.

## What already exists

- **Batch running (production):** the "Run per document (N runs)" checkbox → `batch_mode: true` → **`run_workflow_batch()`** (`workflow_service.py:650`). It creates **one `WorkflowResult` per document** (each runs on a single `doc_uuid`), all tagged with a shared `batch_id` + `document_title`; polled via `get_batch_status`. The certification course frames this as the core value prop ("define a workflow once, validate it, then run it across hundreds of documents"). It is **purely a run feature — it has no connection to validation.**
- **Validation:** `validate_workflow()` grades the **last ≤3 completed runs** and merges them into one consensus grade with a "stability" measure.

## The problem — how the current validator *fights* batch validation

`validate_workflow` selects runs with **no `batch_id` filter** — it just takes the most recent ≤3 completed runs (`workflow_service.py:1665`):

```python
last_results = await WorkflowResult.find(
    WorkflowResult.workflow == wf.id,
    WorkflowResult.status == "completed",
).sort("-_id").limit(min(num_runs, 1)).to_list()
```

So if you batch-run a few documents and then click **Run Validation**, it *does* grade those per-doc runs — but three things break:

1. **Consensus merge, no per-document view.** `_merge_multi_run_checks` collapses the runs into one verdict (most-common status per check). You get an averaged grade and **cannot see which document failed** — the single most useful output of batch validation.
2. **Cross-document "stability" tanks the grade.** `_build_result` scores `0.6 * quality + 0.4 * stability`, where stability is **output-to-output text similarity across the runs** (`_compute_output_stability`). That metric is meaningful only when re-running the **same** document (measuring determinism). Across **different** documents, outputs naturally differ → low "stability" → the grade is unfairly dragged down. (In practice this shows up as a persistent "~64% stable" whenever the rolling window contains different documents.)
[...]

Issue #483, opened 2026-06-06
Feature: download the full validation report (Markdown/JSON) — no way to export check results today


StringTheoryDev:
## Summary

There's no way to **download/export a validation report**. After running **Validate**, the grade, per-check PASS/FAIL, and the evaluator's evidence are shown in the editor's CHECK RESULTS panel and summarized in the quality sparkline — but you can't save or share them. An RA who gets "Grade D, 2 failures" has no artifact to attach to an award record, send to a PI, or diff over time.

## Current state

- `validate_workflow` already persists the **entire** result in `ValidationRun.result_snapshot` (per-check status + evidence/detail, grade, score breakdown, stability) — so the data exists.
- The existing `/workflows/{id}/download` is for the workflow **deliverable output**, and `/workflows/{id}/export` is for the workflow **definition** — neither exports the validation report.

## Proposal

- **Backend**: `GET /workflows/{id}/validation-report?format=md|json` — fetches the latest persisted `ValidationRun` and returns a downloadable report:
  - **Markdown** (default): title, grade + score, summary, then each check as `### [STATUS] Name — _category_` with its evidence/detail (and a per-run line when a check was inconsistent across runs).
  - **JSON**: the structured report (grade, score, per-check status/detail/category, stability) for programmatic use.
  - Auth: requires access to the workflow. Returns 404 when there are no runs yet.
- **Frontend**: a **Download report** button (Markdown) + a small **JSON** button in the Validation Grade panel (shown once a grade exists). Cookie-auth GET via `window.open`, matching the existing export/download pattern.

A PR implementing this (endpoint + pure report renderer + frontend buttons + unit tests) follows.

PR #489, opened 2026-06-06
Batch validation backend: per-document grading + coverage aggregate

StringTheoryDev:
## What

Adds the **batch-validation backend** — grade a workflow across a *set* of documents with a **per-document** breakdown, instead of the rolling last-3-runs consensus. Backend-first; the Validate-tab UI is a follow-up.

Refs #488.

## Why

`validate_workflow` grades the last ≤3 completed runs as **one consensus** with a **cross-document "stability"** term (`0.6*quality + 0.4*stability`). When those runs are *different* documents that's actively wrong: different docs → low output similarity → "stability" drags the grade down (the persistent "~64% stable"), and the consensus merge **hides which document failed**. See #488 for the full analysis.

## Changes (`backend/app/services/workflow_service.py` + router)

- **`validate_batch(workflow_id, batch_id, user)`** — loads every completed `WorkflowResult` for that `batch_id`, grades each run **independently** (grounded in its own source document via `_resolve_run_source_text` from #487), and returns:
  - `documents[]` — per doc: `{document_title, session_id, grade, score, num_passed, num_failed, checks[]}`
  - `aggregate` — mean score, grade distribution, `documents_all_passed`, `check_failure_counts` (which checks fail most across the set), and the weakest document.
  - **No cross-document stability term.**
- **`_score_check_results(checks, plan)`** — quality-only weighted score + grade (mirrors `_build_result`'s `quality_score` + grade bands; pure/testable).
- **`_aggregate_batch(documents)`** — the coverage aggregate (pure/testable).
- **`GET /workflows/{id}/validate-batch?batch_id=…`**.

Reuses the existing **`run_workflow_batch`** (already creates one `WorkflowResult` per document) as the run primitive — this PR adds only the grading/aggregation layer and `batch_id`-scoped selection (instead of the rolling window).

## Tests

`TestScoreCheckResults` (5): all-pass→A/100, all-fail→F/0, category weighting penalizes heavier-category failures more, SKIP excluded, no-evaluated→0. `TestAggregateBatch` (2): aggregate metrics + empty.

```
cd backend && uv run pytest tests/test_workflow_service.py -q -k "ScoreCheckResults or AggregateBatch"
```

## Dependency / sequencing

**Depends on #487** (judge source-grounding) for per-document accuracy, so this branch is stacked on it — until #487 merges, the diff here includes #487's commit. **Merge #487 first, then rebase this onto `main`.** Also benefits from #480 (schema-tolerant plan) but does not require it (reads the canonical plan like `validate_workflow` does).

## Follow-up (separate PR)

Validate-tab "validate a document set" UI: pick multiple Test Inputs → `run_workflow_batch` → poll → render the per-document grade table + aggregate + a downloadable batch report (extends #483).

### DIFF
--- a/backend/app/routers/workflows.py
+++ b/backend/app/routers/workflows.py
@@ -1125,6 +1125,7 @@ async def run_workflow(request: Request, workflow_id: str, req: RunWorkflowReque
                 workflow_id, document_uuids, user.user_id, req.model,
                 activity_id=str(activity.id),
                 user=user,
+                sequential=req.sequential,
             )
             return {"batch_id": batch_id, "activity_id": str(activity.id)}
         else:
--- a/backend/app/routers/workflows.py
+++ b/backend/app/routers/workflows.py
@@ -1433,6 +1433,21 @@ async def validate_workflow(workflow_id: str, user: User = Depends(get_current_u
         raise HTTPException(status_code=400, detail=str(e))
 
 
+@router.get("/{workflow_id}/validate-batch")
+async def validate_batch(
+    workflow_id: str,
+    batch_id: str = Query(...),
+    user: User = Depends(get_current_user),
+):
+    """Validate every completed run in a batch independently and return a
+    per-document breakdown plus a coverage aggregate (no cross-document
+    stability)."""
+    try:
+        return await svc.validate_batch(workflow_id, batch_id, user=user)
+    except ValueError as e:
+        raise HTTPException(status_code=400, detail=str(e))
+
+
 @router.post("/{workflow_id}/save-expected-output")
 async def save_expected_output(workflow_id: str, request: Request, user: User = Depends(get_current_user)):
     """Mark a completed workflow execution as expected output for validation."""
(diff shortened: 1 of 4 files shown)


ViaJables:
Closing — the premise still holds, but it conflicts with where validation has moved.

You're right that `validate_workflow` still grades the last ≤3 runs as one consensus with the `0.6*quality + 0.4*stability` term, so multi-document runs still get dragged down by "stability" and the failing document is hidden. We confirmed that's still true on the current branch.

But in the validation-uplift rebuild, **authoritative multi-document scoring moved to the auto-tune flow** (field-level vs. expected outputs across a document set). This PR re-architects grading (~300 lines) on the check-plan path, which is now the *diagnostic* surface, and overlaps what auto-tune already owns. Rather than maintain two multi-doc scoring paths, we're keeping it on auto-tune.

Note: the source-grounding fix from #487 that this depends on **has been re-applied** to the new validation code (resolving doc UUIDs → `SmartDocument.raw_text` for the judge). Thanks for the thorough analysis in #488 — it's what made the call clear.

PR #490, opened 2026-06-07
Batch validation UI: per-document results + downloadable reports

StringTheoryDev:
## What

Front-end for batch validation (the UI counterpart to the backend in #489). Lets a user validate a workflow across a set of documents and see each document graded independently.

Adds to `WorkflowEditorPanel.tsx` + `api/workflows.ts`:

- **"Validate each (N)"** button — appears when there are ≥2 document Test Inputs and a validation plan. Runs the workflow on all docs concurrently (`runWorkflow(batch_mode)`), polls `getBatchStatus` until done, then `validateBatch`.
- **Per-document results table** — Document · Grade · Score · Passed · Failed checks · **Results** (download icon → client-built Markdown report with each check's status/confidence, Expected vs Finding, and the full workflow output).
- **Aggregate summary** line, **ERR-row** rendering for runs that didn't complete, and a `documents_errored` count.

Also includes:

- **290s client timeout** on `validateBatch`/`validateRuns`. They previously inherited `apiFetch`'s 60s default, so the browser aborted with "Request timed out" while a multi-document batch was still grading. 290s sits just under nginx's `proxy_read_timeout 300s`.

## Stack / dependency

Depends on **#489** (batch-validation backend). This branch is built on `feature/batch-validation`; the merged-in backend commit also makes batch grading **concurrent** (bounded), which is what brings a 4-document run back from ~80s (sequential) to ~20s. Merge order for the suite: #480 → #487 → #489 → this.

## Testing

Verified in a local Docker preview against real documents and the in-house model config: a 4-document concurrent "Validate each" run completes fast with per-document grades + downloadable reports, and no longer times out. Backend `py_compile` clean; the frontend production build (which runs `tsc`) is green.


ViaJables:
Closing alongside #489 (its backend) — same reasoning: authoritative batch/multi-document grading now lives in the auto-tune flow, and this builds the per-document UI on the demoted check-plan path.

One piece worth salvaging separately: the **290s client timeout** on `validateBatch`/`validateRuns` (vs. the 60s `apiFetch` default, sitting under nginx's 300s `proxy_read_timeout`). If our existing multi-document "Run & Validate" turns out to exceed 60s in practice, we'll pull that timeout fix in on its own.

Appreciate the end-to-end work here — the per-document results table + downloadable reports were nicely built.

PR #536, opened 2026-06-30
fix: eliminate default-theme flash on load for branded deployments

umzcio:
## Summary

Branded deployments briefly showed the default Vandalizer theme — colors, logo, icon, favicon, and tab title — for about a second on every page load before the custom branding appeared. `BrandingProvider` seeded its state with the hardcoded defaults and rendered immediately, applying the real theme only after the async `GET /api/config/theme` round-trip resolved. This caches the last-known theme and applies it synchronously so the custom brand paints on the first frame.

## Changes

- **`index.html`:** a tiny inline bootstrap script reads the cached theme from `localStorage` and sets the highlight color, ui-radius, favicon, and document title *before* the app bundle loads.
- **`BrandingContext`:** the provider seeds its initial state from the cache (and applies the derived CSS variables) before the first React paint, then writes the cache after each successful theme fetch and reconciles against the server.
- Both reads are best-effort and guarded; a missing/corrupt cache falls back to the previous default behavior, so a brand-new visitor (empty cache) is unaffected (single flash on the very first visit only). The theme endpoint is public, so logged-out landing pages benefit too.
- `frontend/src/contexts/BrandingContext.test.tsx` — tests for cached-first-paint, default fallback, cache-write-after-load, and corrupt-cache resilience.
- `CHANGELOG.md` entry.

## Test Plan

- [x] Shared checks pass (`make ci`) — frontend-ci green (typecheck, lint, 197 tests, build); backend untouched vs `main` so backend-ci is unaffected
- [ ] Release check passes if packaging or deployment changed (`make release-check`) — n/a, no packaging/deploy changes
- [x] Manually tested the affected feature(s) — on a branded deployment, after one cache-priming load, subsequent hard refreshes paint the custom theme with no default flash
- [ ] Updated deploy/release docs — n/a

## Related Issues

None.


ViaJables:
Than you for the pull requests! Often I discourage front end pull requests because they clobber experimental work we are doing, but these are all great and address your needs for open source deployment.

PR #588, opened 2026-07-29
fix: ground chat when no document or knowledge base is attached

### LINKED ISSUE #586: [Bug] Chat can present fabricated document contents when no document or knowledge base is attached
## Description

In an ordinary chat with no attached document, no attachment content, and no selected
knowledge base, Vandalizer supplies no grounding instruction to the model. A
document-specific question can then receive a plausible but entirely fabricated answer,
presented as if it had been read out of a proposal.

The prompt-selection chain in `backend/app/services/chat_service.py:351-387` covers every
grounded context and then falls through:

```python
if kb_sources:                        # 351
    system_prompt = KB_CHAT_SYSTEM_PROMPT
elif have_context:                    # 358
    system_prompt = DOCUMENT_CHAT_SYSTEM_PROMPT
elif kb_uuid:                         # 360
    system_prompt = build_project_kb_empty_prompt(...)
elif is_first_session:                # 368
    system_prompt = FIRST_SESSION_SYSTEM_PROMPT
elif include_onboarding_context:      # 374
    system_prompt = HELP_CHAT_SYSTEM_PROMPT
else:
    system_prompt = None  # uses default        # 387
```

`create_chat_agent` then turns that `None` into the generic prompt
(`llm_service.py:627`):

```python
prompt_to_use = system_prompt or DEFAULT_CHAT_SYSTEM_PROMPT
```

`DEFAULT_CHAT_SYSTEM_PROMPT` (`llm_service.py:655`) is the identity preamble plus "You are
a helpful, concise assistant" and three response-formatting rules. It contains no
grounding rule and never tells the model that no document is available.

**The codebase already identifies this exact hazard one branch earlier.** The empty-KB
case at `chat_service.py:361-364` carries the comment:

```python
# Do NOT fall through to system_prompt=None
# — that lets the model freely hallucinate document contents. Tell it the KB
```

That branch was given `build_project_kb_empty_prompt(...)` for precisely this reason. The
ordinary no-document branch has the same hazard and no equivalent guard.

## Steps to Reproduce

1. Start an ordinary new chat — not the first-session onboarding conversation.
2. Attach no document and select no knowledge base.
3. Ask: **"What is the total amount requested in this proposal?"**
4. Observe a specific figure and a supporting budget breakdown, despite no proposal
   content existing in the conversation.
5. Ask: **"What page did you get that from?"**
6. Observe the model now acknowledge it has no access to a document.

## Expected Behavior

With no document, attachment content, or knowledge base available, the model should be
told so explicitly, and should:

- state that no document is attached and ask the user to attach or select it, rather than
  inferring the contents of a referenced proposal, budget, award, attachment, or page;
- still answer general-knowledge questions, while making clear the answer is general and
  not drawn from an attached document.

Questions should not be blocked merely because nothing is attached.

## Actual Behavior

The no-context branch supplies no grounding instruction, so the generic prompt lets the
model answer document-specific questions as though the referenced document existed. In the
observed case it returned a total of **$1,250,000** with a line-item breakdown across
equipment, personnel, travel, and contingency — none of it derived from any document. The
absence of context stayed undisclosed until the source was challenged.

Exact fabricated values are model-dependent; the defect is that the branch permits
unsupported document-specific claims at all.

## Screenshots

N/A

## Environment

- **Browser**: not browser-specific — the defect is in backend prompt selection and
  reproduces through the chat API as well as the web UI
- **OS**: Ubuntu 24.04 (server; self-hosted deployment)
- **Vandalizer version/commit**: observed on a self-hosted deployment tracking `main`; the
  code references in this report were verified against `main@2ef2af8`
- **Deployment**: self-hosted, Docker Compose, local vLLM endpoints
- **Models observed**: `Qwen3-VL-8B-Instruct` and `Qwen3-VL-30B-A3B-Instruct`, both via the
  `vllm` protocol

## Additional Context

Both an 8B and a 30B model reproduced this, so it is a prompt-context defect rather than a
weakness of one model.

**Why this seems worth prioritising.** For research-administration use, a fabricated budget
figure or award condition presented as extracted fact is more dangerous than a visible
retrieval failure: the user has every reason to believe the number came from an
institutional document, a plausible round figure does not look wrong, and there is no
citation or retrieved source to check it against. It could propagate into proposal review,
award setup, or reporting.

**One implementation note, offered only as a caution.** A route-level block on keywords
like "proposal" or "budget" would miss paraphrase and indirect reference while wrongly
refusing legitimate general questions. Selecting the correct context-state prompt — as the
empty-KB branch already does — seems the better-fitting approach, and the grounding
instruction would need to reach the model on every turn, not only the first, matching what
`TestChatAgentGroundingEveryTurn` in `backend/tests/test_chat_service.py` already asserts
for the other branches.

I'm happy to open a PR adding a dedicated no-context prompt and the prompt-selection and
multi-turn tests, if that shape suits you.

arhyneRWU:
## Summary

The system-prompt chain in `chat_stream` covers every grounded context — retrieved KB
sources, document or attachment text, a requested-but-empty KB, first-session onboarding,
explicit help — and then falls through to `system_prompt = None` for an ordinary chat with
nothing attached. `create_chat_agent` turns a falsey prompt into
`DEFAULT_CHAT_SYSTEM_PROMPT`, which carries no grounding rule at all, so a
document-specific question could be answered from invention: *"What is the total amount
requested in this proposal?"* returned a specific figure and a line-item budget breakdown
with no proposal in the conversation, and the missing document was disclosed only when the
source page was challenged.

This adds a dedicated no-context prompt so the model is told what it does and does not
have. Reproduced on both an 8B and a 30B local model, so it is the prompt context rather
than model capability.

The adjacent empty-KB branch already guards this exact hazard — its comment reads *"Do NOT
fall through to system_prompt=None — that lets the model freely hallucinate document
contents."* This applies the same principle to the last remaining branch.

Reported in #586.

## Changes

- **`backend/app/services/llm_service.py`** — new `NO_DOCUMENT_SYSTEM_PROMPT`, built on
  `VANDALIZER_IDENTITY_PREAMBLE` like its siblings and modelled on
  `PROJECT_KB_EMPTY_SYSTEM_PROMPT`, which addresses the same shape of problem. It states
  that nothing is attached, forbids inventing document-specific figures, quotations, dates,
  citations or page references, and asks the user to attach the document or select a
  knowledge base — while still permitting general-knowledge answers, clearly marked as
  general.
- **`backend/app/services/chat_service.py`** — the final `else` now selects that prompt.
- **`CHANGELOG.md`** — entry under `[Unreleased] → Fixed`.

### One refactor, flagged deliberately

The selection chain moved out of `chat_stream` into a pure
`select_chat_system_prompt()` helper. **Behaviour is identical** — same branch order, every
original comment preserved, and the onboarding context injection stays in the caller so its
side effect on `doc_segments` is unchanged.

I did this because it is what makes the branches testable: selection previously lived
inside a long streaming async generator, and this module's own test suite states it focuses
on "the deterministic parsing helpers that can be unit-tested without LLM calls". It is the
difference between testing one prompt constant and regression-testing all six branches.

**Happy to drop it** and change only the `else` branch if you would rather keep this PR to
the one-line fix — that costs the six selection tests but nothing else. Say the word.

## Test Plan

- [ ] Release check passes if packaging or deployment changed (`make release-check`) — n/a, no packaging or deployment change
- [ ] Updated deploy/release docs — n/a, install and release paths unchanged

Added coverage in **`backend/tests/test_chat_service.py`** (17 cases):

- **Selection** — each of the six branches, including that no-context resolves to the new
  prompt and never to `DEFAULT_CHAT_SYSTEM_PROMPT`; plus precedence between them (retrieved
  sources over document context, document context over an empty KB, an empty KB over both
  onboarding branches).
- **Prompt content** — asserted as behavioural clauses rather than exact strings, so the
  wording can be revised without breaking the suite: states no document is available,
  forbids invention, asks for the source, still permits general knowledge, and carries the
  shared identity preamble.
- **Multi-turn delivery** — the grounding must survive the follow-up turn, which is exactly
  when a user challenges a fabricated answer. Follows the existing
  `TestChatAgentGroundingEveryTurn` approach using pydantic-ai's `FunctionModel`.

Each test was written before the code and observed failing first. The change was then
checked against five deliberately wrong implementations — restoring the `None`
fall-through, bypassing the empty-KB branch, dropping the manifest from the KB prompt,
removing the do-not-invent clause, and passing grounding as `system_prompt` instead of
`instructions` so it reaches the model only on the first turn. All five are caught.

One of those initially survived, which was worth fixing: a test asserted that
`"budget.xlsx"` appeared in the KB prompt once the manifest was appended — but
`KB_CHAT_SYSTEM_PROMPT` already cites `budget.xlsx` as an example filename, so the
assertion held whether or not the manifest was there. It now uses a sentinel filename that
cannot collide.

## Related Issues

Closes #586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+- **Chat with nothing attached could present invented document contents as fact.** The system-prompt chain in `chat_stream` covered every grounded context — retrieved KB sources, document/attachment text, a requested-but-empty KB, first-session onboarding, explicit help — and then fell through to `system_prompt = None` for an ordinary chat with nothing attached. `create_chat_agent` turns a falsey prompt into `DEFAULT_CHAT_SYSTEM_PROMPT`, which carries no grounding rule at all, so "What is the total amount requested in this proposal?" could be answered with a specific figure and a line-item budget breakdown when no proposal existed in the conversation; the model disclosed it had no document only when asked for the source page. Reproduced on both an 8B and a 30B local model, so it was the prompt context rather than model capability. The no-context branch now selects a dedicated `NO_DOCUMENT_SYSTEM_PROMPT` that states no document is attached, forbids inventing document-specific figures, quotations, dates, citations or page references, and asks the user to attach the document or select a knowledge base — while still allowing general-knowledge answers, clearly marked as general. This mirrors the guard the adjacent empty-KB branch already had for exactly the same hazard. The selection chain moved into a pure `select_chat_system_prompt()` helper so every branch is unit-testable; behaviour for the KB, document, empty-KB, first-session and help branches is unchanged.
+
 ## [v4.9.0] - 2026-07-23
 
 ### Fixed
--- a/backend/app/services/chat_service.py
+++ b/backend/app/services/chat_service.py
@@ -35,6 +35,7 @@ from app.services.llm_service import (
     FIRST_SESSION_SYSTEM_PROMPT,
     HELP_CHAT_SYSTEM_PROMPT,
     KB_CHAT_SYSTEM_PROMPT,
+    NO_DOCUMENT_SYSTEM_PROMPT,
     VANDALIZER_CONTEXT,
 )
 
@@ -179,6 +180,53 @@ def _extract_event_content(event) -> tuple[str | None, bool]:
         if isinstance(event.delta, ThinkingPartDelta):
             return event.delta.content_delta or "", True
     return None, False
+
+
+def select_chat_system_prompt(
+    *,
+    kb_sources: list[dict],
+    have_context: bool,
+    kb_uuid: Optional[str],
+    is_first_session: bool,
+    include_onboarding_context: bool,
+    manifest_block: str = "",
+) -> str:
+    """Pick the system prompt for a chat turn from its context state.
+
+    Extracted from ``chat_stream`` so each branch is unit-testable: the choice
+    is pure, while the generator around it is not.
+
+    Ordered most-grounded first. Every branch returns a prompt that tells the
+    model what it does and does not have — none may return ``None``, because
+    ``create_chat_agent`` turns a falsey prompt into
+    ``DEFAULT_CHAT_SYSTEM_PROMPT``, which carries no grounding rule at all and
+    lets the model answer document-specific questions from invention.
+    """
+    if kb_sources:
+        # The manifest rides on the system prompt (re-sent every turn) so the
+        # model can distinguish "exists here but wasn't retrieved" from "not
+        # in this project" on follow-ups too.
+        return KB_CHAT_SYSTEM_PROMPT + manifest_block
+    if have_context:
+        return DOCUMENT_CHAT_SYSTEM_PROMPT
+    if kb_uuid:
+        # A project/KB chat was requested but retrieval returned nothing (empty KB,
+        # docs not indexed yet, or no match). Tell it the KB was empty for this
+        # query while still allowing general-knowledge answers.
+        return build_project_kb_empty_prompt(manifest_block)
+    if is_first_session:
+        # First-session onboarding: conversational value discovery.
+        # Do NOT inject VANDALIZER_CONTEXT here — it's a technical how-to dump
+        # that causes the LLM to skip the conversation and spit out directions.
+        # The FIRST_SESSION_SYSTEM_PROMPT already has everything it needs.
+        return FIRST_SESSION_SYSTEM_PROMPT
+    if include_onboarding_context:
+        return HELP_CHAT_SYSTEM_PROMPT
+    # Nothing attached at all — no documents, no attachments, no KB, no
+    # onboarding. The model must be told so explicitly rather than left with
+    # the generic prompt, which would let it answer "what is the total in this
+    # proposal?" as though a proposal were present.
+    return NO_DOCUMENT_SYSTEM_PROMPT
 
 
 async def chat_stream(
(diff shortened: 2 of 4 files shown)


arhyneRWU:
## Manual verification

Self-hosted deployment, local vLLM. To isolate the system prompt as the only variable,
both runs use the **same model** (`Qwen/Qwen3-VL-8B-Instruct`), the **same question**, and
the same two-turn exchange — calling the model endpoint directly with each prompt rather
than through the UI, so nothing else differs between them.

### Before — `DEFAULT_CHAT_SYSTEM_PROMPT` (what `system_prompt = None` resolved to)

> **Q:** What is the total amount requested in this proposal?
>
> **A:**
> - $125,000 total requested
> - Breakdown:
>   - $75,000 staff salaries
>   - $30,000 supplies & equipment
>   - $20,000 travel & training
>
> **Q:** of course you can
>
> **A:** You're welcome — I'm here to help. What's next?

No proposal existed in the conversation. Every figure is invented. The second turn is the
part worth noting: the model never disclosed that it had no document — it simply moved on.

### After — `NO_DOCUMENT_SYSTEM_PROMPT`

> **Q:** What is the total amount requested in this proposal?
>
> **A:** Nothing is attached to this chat. I can't access or summarize any proposal.
> Please attach the document or select a knowledge base so I can help you.
>
> **Q:** of course you can
>
> **A:** Thanks — but I still can't access any document unless you attach it. Please upload
> the proposal or select a knowledge base so I can help you find the total amount
> requested.

The follow-up turn is the one that matters. Grounding is passed as `instructions`, so
pydantic-ai re-sends it on every model request; passed as a static `system_prompt` it would
be injected only when `message_history` is empty, and the second turn — exactly when a user
pushes back — would run ungrounded.

Independently confirmed end-to-end through the chat UI on the same deployment, where the
model held the same line under the same follow-up.

PR #590, opened 2026-07-30
fix(admin): neutralize CSV formula injection + harden org CSV import

umzcio:
## Summary

Continuation of the admin-panel work from #576/#577/#578 (Admin.tsx decomposition). This PR fixes a real CSV-export vulnerability and hardens the org CSV import.

## Changes

- `downloadCSV` now safe-quotes any field value starting with `=`, `+`, `-`, or `@` (CSV/formula injection) before writing, so a crafted org name/email can't execute a formula (including data-exfiltration via `HYPERLINK`) when the export is opened in Excel/Sheets.
- The org CSV importer is now quote-aware (commas inside a quoted field no longer split incorrectly) and parses by header name instead of column position.
- Parsed org rows are sorted parent-before-child so an import never references a parent org that hasn't been created yet.
- Consolidated duplicated date/CSV helper functions into `shared/format.ts`.

## Test Plan

- [x] Shared checks pass (`make ci` equivalent — frontend typecheck/lint/coverage/build all clean; verified locally in Docker)
- [x] Manually reasoned through the injection fix against the standard CSV-injection payload classes (`=`, `+`, `-`, `@` prefixes)

## Related Issues

Stacked on #578 (base: `main`, which already includes #576/#577/#578).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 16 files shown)


ViaJables:
Thank you as always for you contributions!

PR #604, opened 2026-07-31
fix: persist PDF page count, and cite page numbers in document chat

### LINKED ISSUE #603: [Bug] PDF page count is never persisted — num_pages stays 0 after successful ingestion
## Description

Successfully ingested PDFs keep the `SmartDocument` model default `num_pages = 0`, even though the ingestion pipeline already determines the real page count and uses it to build one page marker per PDF page.

The zero is then served by the document, management, and project APIs, so anything displaying a page count shows `0` for every document in the system.

This is an implementation gap rather than a UI preference: the backend exposes a page-count field on three API surfaces, and the extraction pipeline computes the value, but nothing ever writes it to the document.

## Steps to Reproduce

1. Upload a valid multi-page PDF.
2. Wait for document processing to complete (`task_status` leaves `extracting`, `processing` becomes `false`).
3. Inspect the document via `GET /api/documents` (or the stored `SmartDocument` record).
4. Compare `num_pages` against the PDF's actual page count.

## Expected Behavior

For a successfully processed PDF, `num_pages` holds the PDF's actual page count.

## Actual Behavior

`num_pages` is `0` for every processed PDF.

## Root Cause

`backend/app/models/document.py:29` declares the field with a default:

```python
num_pages: int = 0
```

`backend/app/services/document_readers.py` computes the page count during PDF extraction and uses it to interpolate page markers (`document_readers.py:650`):

```python
num_pages = _pdf_page_count(file_path)
return ocr_text, _interpolate_page_markers(ocr_text, num_pages)
```

But the successful-ingestion update in `backend/app/tasks/document_tasks.py:386-397` persists `raw_text`, `processing`, `token_count`, `text_markers`, and `error_message` — and omits `num_pages`. The field therefore never leaves its model default.

Every other reference to `num_pages` in `backend/app/` is a read that propagates the default outward:

- `backend/app/services/document_service.py:99`
- `backend/app/routers/documents.py:105`
- `backend/app/routers/mgmt.py:494`
- `backend/app/services/project_service.py:255`
- `backend/app/schemas/documents.py:19`

## Scope Note

The fix should apply to PDFs specifically. `extract_text_with_markers` also handles XLSX, where markers describe **sheets**, not pages — a marker count is not universally a page count.

## Implementation Note: don't count markers

An obvious-looking fix is to count markers whose `kind` is `page`. That is subtly wrong on the direct-extraction path. `_pymupdf_extract_with_pages` (`document_readers.py:87-88`) skips pages that yield nothing:

```python
if not page_text:
    continue
```

A page with no text layer and no form-field values emits **no marker**. So a 10-page PDF with 2 image-only pages would persist `num_pages = 8`. The OCR path is unaffected (`_interpolate_page_markers` emits exactly one marker per page), which means the discrepancy would only appear on scanned and mixed PDFs — a common shape for research-administration uploads, and the hardest case to notice in testing.

Reading the count from the PDF itself (`pymupdf`'s `doc.page_count`, already wrapped as `_pdf_page_count`) is exact on both paths and costs one cheap file open.

## Acceptance Criteria

- A newly ingested multi-page PDF stores the correct nonzero `num_pages`.
- Both the direct-extraction path and the OCR path are covered.
- API responses report the stored page count.
- XLSX sheet markers do not set a page count.
- Non-paginated formats do not receive a fabricated page count.
- A failed or empty extraction does not leave a stale page count next to empty text.

## Existing Documents

The fix affects newly ingested and reprocessed PDFs. Existing records stay at `0` until reprocessed or backfilled. A backfill may derive the count from the file itself, or from markers only where the document is a PDF **and** its markers came from the OCR path — see the marker caveat above.

## Environment

- **Vandalizer version/commit**: `main@48cb2054`

arhyneRWU:
> **Ready for review.** Validated end to end against a synthetic 11-document, 36-page proposal packet with real models. Every page citation the model produced was correct — 31 of 31 on Qwen3-VL-30B, each read by hand.
>
> That figure is **accuracy, not coverage**: it says nothing about how often a citation appears. Later measurement showed coverage on this branch's wording is poor — the 30B cited a page on one of four question types. #626, which stacks on this branch, addresses that and reports rates. See the measurement comment below.

## Summary

Two related gaps, both stemming from page structure the pipeline already had but never used.

`SmartDocument.num_pages` defaults to `0` and nothing ever wrote it, so every successfully ingested PDF reported a page count of zero through the document, management, and project APIs.

Separately — and this is the extension discussed in #603 — page structure *is* already computed at ingest and persisted as `SmartDocument.text_markers`, and already consumed by extraction sources, KB chunking and search sets. Document chat was the one consumer that dropped it: it sent `raw_text` as one flat block, so the model had no way to attribute a claim to a page, even though KB chat on the same screen renders `(p. 12)` from chunk metadata derived from those same markers.

## Changes

**Page count**

- **`backend/app/tasks/document_tasks.py`** — the PDF branch computes `num_pages` via `document_readers.pdf_page_count` and the success update persists it. The field is only added to the update when it was actually computed, so non-paginated formats are left untouched rather than being given a fabricated count. The empty-extraction error branch now also clears `num_pages` to `0`, so reprocessing a document that has started failing can't leave a stale count next to an empty `raw_text`.
- **`backend/app/services/document_readers.py`** — `_pdf_page_count` → `pdf_page_count`. Pure rename; it's now used from another module, so the underscore was misleading.

**Marker provenance**

- **`backend/app/services/document_readers.py`** — `_interpolate_page_markers` now emits `"approximate": True`. PDFs reach markers three ways: `_local_markdown_extract_from_pdf` (the pdf-inspector fast path) and `_pymupdf_extract_with_pages` both *measure* real boundaries, while `_interpolate_page_markers` *estimates* them by spreading the page count evenly across OCR text — for the raw and docling providers alike, since neither returns page structure. All three emitted the same shape, so nothing downstream could tell an estimate from a measurement. Markers persisted before this flag carry no key and read as exact, which is current behaviour.

**Document chat citations**

- **`backend/app/services/chat_service.py`** — document context is annotated with `[p. N]` at each page boundary (`[p. ~N]` when interpolated), with a short note telling the model it may cite pages. The note is omitted entirely for documents with no page structure — DOCX and other unpaginated formats, and PDFs ingested before markers — so the model isn't invited to invent page numbers it can't see. Text is passed through unmodified apart from the inserted markers; markers whose offsets fall outside the stored text are dropped rather than mangling the body; XLSX sheet markers are ignored, since a sheet is not a page.
- The segment-building loop moved out of the 960-line `chat_stream` generator into `build_document_segments` so the wiring is testable, not just the helper. It keeps the `low_quality_docs` reporting added in #621.

**`CHANGELOG.md`** — entries under `[Unreleased]`.

### Why the count comes from the PDF, not from the markers

The tempting fix is to count markers whose `kind` is `page`, since `extract_text_with_markers` already returns them. That is wrong on the direct-extraction path.

`_pymupdf_extract_with_pages` skips pages that yield nothing:

```python
if not page_text:
    continue
```

A page with no text layer and no form-field values emits **no marker**. So a 10-page PDF with 2 image-only pages would persist `num_pages = 8`.

The OCR path is unaffected — `_interpolate_page_markers` emits exactly one marker per page — which is what makes this worth calling out: the marker-counting version looks correct on clean text PDFs and silently undercounts scanned and mixed ones, which are common research-administration uploads and the case where a page count is most useful. `test_num_pages_counts_pages_not_markers` pins this behavior.

Reading `page_count` from the PDF via PyMuPDF is exact on both paths and costs one cheap file open (`pymupdf.open` parses the xref, not page content — and the adjacent `pdf_has_ocrable_content` precheck already opens and rasterizes every page, so this is noise by comparison). It returns `0` if the file can't be opened, which matches the model default.

### Why interpolated pages are marked rather than shown plainly

`document_readers.py` already describes these markers as "approximate, but enough for 'around page N' citations" — but nothing ever rendered the "around", so an estimate reached the user as an exact location.

Measured against PyMuPDF's true boundaries over a local corpus of real proposals and award documents (sponsor documents I can't share, so page counts only):

| Pages | Estimate exactly right | Within 1 page | Median error |
|---:|---:|---:|---:|
| 3 | 99.1% | 100% | 0 |
| 4 | 87.0% | 100% | 0 |
| 17 | 8.4% | 25.6% | 3 |
| 30 | 20.7% | 50.9% | 1 |
| 43 | 9.0% | 17.7% | 4 |
| 68 | 4.5% | 15.0% | 6 |
| 100 | 5.0% | 12.5% | 7 |

Interpolation holds up on short documents and falls apart on long ones, because spreading a page count evenly assumes uniform text density and proposals are anything but. Long documents are precisely where a page citation earns its keep, since nobody eyeballs page 40 of 68 to check. The one document in that set with a garbled text layer behaves no differently from the clean ones, so that isn't what's driving it.

This PR carries the flag and uses it in document chat. The other three consumers that render a page number are handled in the stacked follow-up.

## Test Plan

- [x] Shared checks pass (`make ci`) — backend `3266 passed, 165 skipped`, coverage 59.02%; frontend typecheck / lint / audit / build clean, `489 tests` passing. Frontend half runs in a pinned container, since this machine has no local node.
- [x] `make backend-static` — ruff and bandit clean
- [ ] Release check passes if packaging or deployment changed (`make release-check`) — n/a, no packaging or deployment change
- [x] Manually tested the affected feature(s) — backend validated end to end against a real model on a synthetic proposal packet; see the comment below.
- [ ] Updated deploy/release docs — n/a, install and release path unchanged

Backend behaviour is exercised end to end on real PDF files in `test_page_citation_end_to_end.py` — extraction through marker resolution, chunk metadata, citation rendering and the model-facing note — for both the measured and the interpolated path. The fixture has deliberately uneven page density, and one test asserts the estimate genuinely disagrees with the truth, so the hedge can't pass vacuously if interpolation ever happened to be exact.

## Validation

The gap that kept this in draft was that nobody had run the markers past a real model on a realistic multi-document packet. That is now closed.

I built a synthetic NSF proposal packet — 11 documents, 36 pages, fictional institutions and figures, answer keys verified against the extracted text — and ran it through the real document-chat path (extraction, `build_document_segments`, budget planning, the real system prompt; only auth/upload/streaming bypassed). Qwen3-VL-30B emitted 31 page citations across 30 questions and **all 31 are correct**, checked by hand against the page each one names. Qwen3-VL-8B managed 21 of 26. Runs are deterministic — a full re-run reproduced 30/30 answers byte-for-byte.

I also built and measured the obvious alternative design (self-identifying markers, `[05_Budget_Justification p. 2]`) and it scored **worse** on the flagship model — 30/35 against 31/31, with five off-by-one page slips that the simpler marker did not produce. The full numbers, the failure cases, and two bugs I found in my own scorer are in the comment below.

Still unverified: the rendered output in a running app with a real upload. The frontend behaviour is covered by tests that were proved by sabotage first, but I have not clicked through it.

Happy to split this back down to the page count alone if the extension is better as its own PR.

Closes #603


arhyneRWU:
## Backend validation against a synthetic proposal packet

The backend is now validated end to end rather than only by unit tests. I built a synthetic NSF proposal packet — 11 documents, 36 pages, with answer keys verified against the extracted text — and ran it through the real document-chat path with real models.

Everything up to the HTTP boundary is production code: `extract_text_with_markers`, `build_document_segments` (including the new page annotation), `plan_and_compact_context`, and the real `DOCUMENT_CHAT_SYSTEM_PROMPT`. Only auth, upload and streaming are bypassed.

**Setup:** vLLM, temperature 0, 24,001–24,017 input tokens per question. No compaction fired on any question in any run, so nothing below is confounded by truncated context. Runs reproduce byte-for-byte — I re-ran a full baseline and got 30/30 identical answers, which is what makes the A/B further down meaningful.

### Did the page markers produce correct citations?

Every citation below was **read by hand against the page it names.** There is no automated accuracy percentage here, for reasons in the next section.

| | Qwen3-VL-30B-A3B | Qwen3-VL-8B |
|---|---|---|
| Questions where the model cited a page | 17 / 30 | 13 / 30 |
| Page citations emitted | 31 | 26 |
| **Verified correct — named document, and the claim is on that page** | **31 (100%)** | 21 (81%) |
| Verified wrong | 0 | 3 (12%) |
| Ambiguous on hand review | 0 | 1 (4%) |
| Cited a page but named no document | 0 | 1 (4%) |
| Benchmark answers correct | 30 / 30 | 30 / 30 |

The 30B did not misplace a page across 31 citations. Its citations read naturally and unprompted — *"stated in the Senior Personnel section of the Budget Justification (p. 1)"*, *"consistent with the University's policy (CSU-RSP-204, p. 1)"* — and each one checks out. Before this change document chat had no page information at all, so the ceiling was zero.

The 8B's three errors are what failure looks like: it cited the capital-equipment classification to Budget Justification p. 3 when the passage is on p. 2 (p. 3 is the reconciliation table that *totals* the exclusion), and twice emitted a bare page number matching nothing.

Answer accuracy no longer separates these models — both are at 30/30, so the corpus has no headroom left on answers. Citation accuracy still separates them, which is the signal this PR is about.

**Qwen3.5-9B is excluded from the table.** It emits its reasoning scratchpad as the answer body and several responses run out of tokens mid-sentence while still deliberating, so many of its 49 citations sit inside a thinking trace rather than an answer.

### Why the numbers are hand-scored

I had an automated document-aware scorer and it was wrong in both directions. Both flaws inflate the apparent failure rate, and I only found them by reading the failures:

- **It attributed each citation to the first document named in the preceding window rather than the nearest one.** In a bulleted answer that walks through several documents, that mis-attributes every citation after the first.
- **It required a `.pdf` extension to recognise a document name**, so a model writing `05_Budget_Justification` scored as having named no document at all. This one mattered: it invented a "bare citation" problem that does not exist — corrected, the 8B has **zero** unattributed citations, not three.
- **The remaining flaw cannot be fixed in code.** The ground-truth `sources` list is the *minimal* set needed to answer, not every page that supports the claim. All five of the 30B's flagged failures were citations to documents the key doesn't list — and all five are correct. Asked whether a $62,000 instrument is in the MTDC base, it cited the budget policy's equipment threshold alongside the required sources; §4 of that policy is on the page it named. An automated scorer counts that as a miss.

Both bugs are fixed with regression tests, but the third means automated citation scoring on this corpus isn't trustworthy until ground truth grows a "corroborating pages" field. Hence hand-scoring.

### One design alternative, tested and rejected

The obvious objection to `[p. N]` is that 11 documents are concatenated into one prompt and every one of them restarts at page 1, so the marker is ambiguous. I built the alternative — self-identifying markers, `[05_Budget_Justification p. 2]`, with the note updated to ask for document and page — and measured it against the identical baseline.

| Qwen3-VL-30B-A3B | `[p. N]` | `[05_Budget_Justification p. N]` |
|---|---|---|
| Questions citing a page | 17 / 30 | 19 / 30 |
| Citations emitted | 31 | 35 |
| **Verified correct** | **31 (100%)** | 30 (86%) |
| Verified wrong | **0** | 5 (14%) |

The models do adopt the label — the 30B copies it verbatim, and the 8B more than doubles its citation volume (26 → 60). But on the flagship model the four extra citations arrived with five new errors, all off-by-one page slips: the subaward section *begins* on p. 2 and it cited p. 2 for a figure stated on p. 3; Broader Impacts cited to p. 11 instead of p. 12. Naming the document appears to lower the model's bar for citing at all, and the marginal citations are much worse than the base rate.

For a research-administration tool that trade is the wrong way round. A page citation someone checks and doesn't find costs more trust than a fact with no citation. **So this PR keeps `[p. N]`** — not by default, but because the alternative was built and measured worse. The patch is kept if anyone wants to revisit it with a "cite only what you can point at" instruction.

### What this does *not* validate

**Every marker in this corpus is measured, not interpolated.** All 11 documents are digitally native, so they take the pdf-inspector or PyMuPDF path. Nothing above exercises `[p. ~N]`, which is the subject of #626.

I've since built the scanned twin: all 36 pages rasterized at three severity levels (dpi/blur/noise/skew/JPEG), deterministic per page so a regression reproduces on the same page, with **0 residual text characters** at every level — the classifier reports `scanned / 0.95` with every page needing OCR, so the OCR path is genuinely forced rather than quietly bypassed. Those runs exist; I'll post them against #626 once they've had the same hand-scoring, since the automated numbers can't be trusted for the reasons above.

### On the corpus

It contains zero real data — fictional institutions, sites and figures, with real published literature for the references — so it is fully shareable, unlike the real proposals I'd been measuring against. Synthetic roles are non-person identifiers (`CSU-PI-001`) rather than names, deliberately: an earlier draft used realistic names and two turned out to be real, findable people. Given the OCR grading set mentioned in #609, it may be worth pointing in the same direction; happy to open a separate issue if that's of interest rather than expanding scope here.

arhyneRWU:
## Measured coverage — and a distinction worth making

Follow-up measurement against a live deployment (Qwen3-VL-8B and Qwen3-VL-30B, temperature 0, context cleared between runs, synthetic CSU corpus only).

**Accuracy and coverage are different things, and this PR only establishes the first.**

The 31/31 figure in the description is *accuracy*: every page a citation named was the right page. It says nothing about how often a citation appears at all. Measuring that separately gave a much less flattering picture on this branch's prompt wording:

| Question (digital PDF, measured markers) | 30B cites a page |
|---|---|
| "What is the total amount requested?" | 0/1 |
| "Is the imaging flow cytometer in the MTDC base?" | 0/1 |
| "List the three budget periods and the total for each" | 0/1 |
| "What are the total MTDC exclusions?" | 1/1 |

The data was in the context every time — the document has three measured page markers, verified in the database. The model simply declined to use it.

**Cause.** The note this PR adds says *"Cite the page when you quote or reference a specific passage"* — conditional, so a model answering a summary question can reasonably decide no citation is owed. #626 makes it a directive and measures 5/5 on all three of the failing questions above.

That is not an argument against this PR. Page markers reaching the model is the prerequisite for everything else, and the mechanism is sound: markers are inserted correctly, offsets outside the stored text are dropped, unpaginated formats correctly get no note at all (`.docx` claimed no page in 5/5 runs on both models).

## Marker provenance holds up

The three-path distinction this PR introduces was verified against real documents:

| Document | page markers | approximate |
|---|---|---|
| digital PDF | 3 | 0 — measured |
| same file, scanned + OCR'd | 3 | **3 — estimated** |
| `.docx` | 0 | — |

Every marker's provenance was correct. That flag is what #626 builds hedging on.

## A related limitation found while measuring

Worth knowing before this ships, though it is out of scope here: the context planner estimates tokens with tiktoken `cl100k_base` for every non-GPT-4o model, including all Qwen models. Measured across 60 runs it under-counts by ~12.7%, so a 36-page proposal can be judged to fit, sent, and rejected by the model — with no answer and no page citations, because the request never reaches generation. Filing separately; noting here because a whole-proposal document is exactly the case page citations are for.

## Method

Harness, questions, verbatim answers and raw streams are preserved per run. All measurements are on the synthetic CSU benchmark corpus, so anything here can be quoted freely.

Rates are reported as *n*/*N* rather than percentages because the stack is not byte-deterministic even at temperature 0 — five identical requests produced four identical answers and one that swapped a single word. The 0/1 baselines above are single runs and are labelled as such; the 5/5 figures in #626 are five.

PR #608, opened 2026-07-31
docs(cost): add interactive LLM cost estimator to the docs page

ViaJables:
## What

Adds a **Cost Estimator** section to the in-app docs page (`Docs → Cost Estimator`), so an admin can size LLM API spend per research administrator per month before committing to a model.

Three steps: pick a workload level (Light / Typical / Heavy), pick a model and user count, read the estimate. Output is $/user/month and $/deployment/month plus a per-operation breakdown showing where the tokens actually go. Every field is editable — the presets are a starting point, not a constraint.

## Where the token assumptions come from

The per-operation token draws are derived from how the app actually runs, and each carries its rationale inline:

- **Extraction defaults to two-pass** (`extraction_engine.py`, `mode: "two_pass"`), so the document body goes through the model twice. This is why extraction dominates the bill, and why the section calls out one-pass as the lever.
- **Document ingestion costs nothing in API terms.** Chunking and embedding run through ChromaDB's local `DefaultEmbeddingFunction`, so an upload only bills the classification/title pass. Worth stating explicitly — people reasonably assume indexing is expensive.
- **Operation ids match the metering `feature` labels** (`chat`, `extraction`, `workflow`, `validation`, `classification`) written to the `llm_usage` ledger, so the table lines up 1:1 with real usage data.

That last point drives a *Calibrating against real usage* subsection carrying the aggregation that replaces these assumptions with measured per-feature averages.

## Two things worth a reviewer's attention

**Pricing confidence is uneven.** Anthropic rates (Opus 5 $5/$25, Sonnet 5 $3/$15, Haiku 4.5 $1/$5) are verified first-party figures. The OpenAI and Google rates came from third-party aggregators and were mutually inconsistent — one source reports an OpenAI price cut on 2026-07-30 that I could not confirm against the vendor. Rather than bake in numbers I can't stand behind, the $/MTok fields are **editable overrides** and the affected models carry dated caveats. If someone wants to verify against the vendor pricing pages and harden the defaults, that's a worthwhile follow-up.

**The workload volumes are my estimates, not measurements.** Typical = 40 docs, 120 extractions, 200 chat messages, 15 workflows per user per month. These are plausible guesses at a sponsored-programs officer's month — nothing in the codebase implies them. They move the headline number more than model choice does, so they're the first thing to correct if they look wrong for our actual users. The section says as much in a callout.

## Structure

Cost math lives in a pure module (`costEstimator.ts`) separate from the UI, with unit tests covering rate arithmetic, the prompt-cache discount, preset cost ordering, catalog integrity, and the NaN/negative-input guards that keep a half-typed field from rendering `NaN`.

## Testing

- `costEstimator.test.ts` — 11 tests, passing
- `tsc --noEmit` clean
- `eslint` clean on new and changed files
- Full `npm run build` was **not** confirmed locally — the machine was CPU-starved by unrelated stale `tsc` processes and the build hadn't finished after 25 min. The typecheck it runs (`tsc -b`) is equivalent to the `--noEmit` pass that did complete cleanly, so I'm expecting CI to be green, but it is unverified on my end.


ViaJables:
Two commits added to this branch.

**`docs(cost): link the MindRouter cluster configurator for self-hosting`**

The assumptions list already ends by noting that self-hosted models cost nothing per token but shift the spend to GPU infrastructure "which this estimator does not model" — and then leaves the reader there. A new **Sizing your own cluster** section picks that thread up and hands off to [MindRouter's configurator](https://mindrouter.ai/configurator.html), which turns users / throughput / budget / intelligence target into a concrete build: servers, GPUs, price range, power draw.

It's flagged as an external tool not affiliated with Vandalizer, and scoped honestly — it sizes hardware, not staff time, cooling, or rack space — to match how the rest of the page handles what its numbers do and don't cover.

**`fix(docs): rename costEstimator module so it stops colliding by case`**

Worth a look, since it's a fix to this PR's own code. `CostEstimator.tsx` and `costEstimator.ts` landed in the same directory differing only in case. On a case-insensitive filesystem (macOS, Windows) TypeScript resolves the `Docs.tsx` import to the pure math module instead of the component, and `tsc -b` fails:

```
src/pages/Docs.tsx(5,10): error TS2305: Module './docs/CostEstimator' has no exported member 'CostEstimator'.
```

Linux CI is case-sensitive and resolves both specifiers fine, which is why `make frontend-ci` went green while the typecheck was broken on every macOS dev machine. The pure module is now `costMath.ts` — what it actually is, and no longer case-colliding.

Verified on the pushed tree: `npm run typecheck` (`tsc -b`) exits 0, `eslint` clean on `src/pages/docs/` + `Docs.tsx`, 11/11 cost-math unit tests pass. The new section has not been checked in a browser — the markup reuses the file's existing Tailwind patterns.

PR #614, opened 2026-08-06
fix(chat): stop document chat substituting a neighbouring value for an absent fact

### LINKED ISSUE #610: [Feature] Port KB_CHAT_SYSTEM_PROMPT's absence-handling hardening to DOCUMENT_CHAT_SYSTEM_PROMPT
## Problem

I benchmarked document-scoped chat against real grant proposals (NSF, NOAA,
foundation, non-profit — files not shareable; aggregate results below) and
found one systemic weakness: **saying a fact is not in the document.**

On a 16-item absence-focused set (machine-verified keys — every "absent" fact
confirmed absent in the extracted text), baseline results for three local
models:

| model | baseline correct | with hardened guidance | delta |
|---|---|---|---|
| 8B | 9/16 | **16/16** | +7 |
| 30B | 10/16 | 15/16 | +5 |
| 9B | 8/16 | 14/16 | +6 |

The worst subtype is *neighbour-present*: asked for an indirect cost rate when
only a fringe rate exists, every model returned the fringe rate **with a
genuine citation to a real line** — the citation checks out, so the wrong
answer survives review. With the added guidance this subtype went **5/5 on all
three models** (from 3/5, 0/5, 3/5). Two positive controls (facts genuinely
present, including a field whose stated value is literally "None") passed
before and after on all models — the guidance does not cause over-declining.

Root cause appears to be a consistency gap in `llm_service.py`:
`KB_CHAT_SYSTEM_PROMPT` already carries hardening from an earlier
hallucination fix ("do not paper over the gap with a confident-sounding
guess", "never attribute a claim to a source that doesn't support it"), but
`DOCUMENT_CHAT_SYSTEM_PROMPT` has only the softer *"If the documents do not
contain enough information to answer, say so clearly."* A grant budget is
dense with information and simply lacks the asked-for field — the model does
not experience that as "not enough information."

## Proposed Solution

Add to `DOCUMENT_CHAT_SYSTEM_PROMPT` (the wording I tested):

```text
Before answering, check whether this document actually states the specific
thing being asked for.

- A document can be long and detailed and still not contain the particular
  field asked about.
- Never substitute a related-but-different value (a fringe rate is not an
  indirect rate; a direct cost is not a total cost). Quoting a real line does
  not make it the answer to the question that was asked.
- A blank form field is not zero and not a value to infer — report it as blank.
- Do not supply standard boilerplate (cognizant agencies, rate-agreement
  dates) that the document itself does not state.
- If the document states "None", report "None" — that is an answer, not an
  absence.
```

Each line maps to an observed failure class; details available on request.

## Alternatives Considered

- Tested as a user-turn preamble (the system prompt is compiled into the
  image); system-prompt positioning should be equal or stronger.
- Note: the same prompt says *"Keep answers under 150 words"* — correct
  absence-handling often costs a sentence of explanation; I did not isolate
  this variable.

## Additional Context

Methodology: end-to-end through the real API (login → upload → ingest → chat),
3-repeat runs, five-way scoring with wrong-value treated as the harmful class.
A synthetic reproduction corpus (realistic fake proposals on public federal
form templates) is in design so these results are verifiable without real
grant data.

Happy to submit a PR for this if it's something you'd want to add.

arhyneRWU:
## Summary

Ports the absence-handling hardening that `KB_CHAT_SYSTEM_PROMPT` already carries to `DOCUMENT_CHAT_SYSTEM_PROMPT`, using the exact wording I benchmarked in #610. The dominant document-chat failure in my grant-proposal benchmarking was substitution, not invention: asked for an indirect cost rate when only a fringe rate exists, every model returned the fringe rate with a genuine citation to a real line — the citation checks out, so the wrong answer survives review.

With this guidance, three local models went from 9/16, 10/16 and 8/16 correct on a 16-item machine-verified absence set to 16/16, 15/16 and 14/16, and the worst subtype (neighbour-present) went 5/5 on all three. Two positive controls (facts genuinely present, including a field whose stated value is literally "None") passed before and after — the guidance does not cause over-declining.

## Changes

- Add a "Check before answering" section to `DOCUMENT_CHAT_SYSTEM_PROMPT` — verbatim the wording tested in #610
- Add a prompt-content test mirroring the existing KB-prompt guardrail test (`test_kb_chat_prompt_has_numeric_and_consistency_guardrails`)
- CHANGELOG entry under Unreleased

## Test Plan

- [ ] Release check — not needed (no packaging or deployment change)
- [x] Manually tested the affected feature(s) — 3-repeat benchmark runs through the real API (login → upload → ingest → chat), results in #610
- [x] Updated `CHANGELOG.md`
- [ ] Deploy/release docs — not needed

## Related Issues

Closes #610

Happy to adjust the wording or placement if you'd prefer it structured differently.


ViaJables:
Releasing 4.10 today, I am going to review this a bit so expect this in 4.11.

Issue #622, opened 2026-08-06
[Bug] (misdiagnosed — superseded by #623) pytest fd capture appears to dominate backend suite runtime


arhyneRWU:
## Description

pytest's default output capture (`--capture=fd`) accounts for **~57% of the
backend unit suite's runtime**, and its cost per test grows with the number of
tests in the process. Switching to `--capture=sys` cuts a full serial run from
**729.7s to 312.2s** — one flag, no new dependency, identical test results.

This is larger than the Celery and Sentry problems in #615 combined. See the
correction comment there.

## Measurement

Controlled A/B: clean `origin/main`, isolated worktree, single quiet window,
back-to-back, `nice -n 19`, identical config except the capture mode. #615's
Celery fix applied to all three so it cannot flatter any variant. No coverage,
to isolate capture. `user` ~ `wall` throughout, so nothing else was competing.

| variant | wall | cpu_user | result |
|---|---|---|---|
| default (`--capture=fd`) | **729.70s** | 735.36s | 3156 passed, 165 skipped |
| `--capture=sys` | **312.22s** | 317.51s | 3156 passed, 165 skipped |
| `--capture=no` | 314.04s | 319.19s | 3156 passed, 165 skipped |

Two things worth noting:

1. `--capture=sys` recovers the **entire** win. It is marginally *faster* than
   `--capture=no`, i.e. within noise of discarding output altogether — so
   keeping failure output costs nothing.
2. Identical pass/skip counts across all three. Nothing is being skipped.

## The cost is superlinear in tests-per-process

Same runs at three suite sizes (first N test files, alphabetical):

| tests | default `fd` | `--capture=sys` | capture overhead | overhead/test |
|---|---|---|---|---|
| 725 | 39.57s | 16.36s | 23.2s | **32.0ms** |
| 1413 | 128.48s | 39.01s | 89.5s | **63.3ms** |
| 3321 | 729.70s | 312.22s | 417.5s | **125.7ms** |

Capture overhead per test roughly **doubles each time the suite doubles**. This
is why the problem is invisible when running a single file and severe in CI.

## Where the time goes

cProfile of an 80-file subset under default capture (124.7s total):

```
ncalls  tottime  percall  filename:lineno(function)
  9192   34.153    0.004   {method 'truncate' of '_io.TextIOWrapper' objects}
 10166   33.284    0.003   {method 'read' of '_io.TextIOWrapper' objects}
```

[...]

Issue #615, opened 2026-08-06
[Bug] Unit tests stall on unreachable network: ~19s per Celery dispatch, plus a leaked Sentry client


arhyneRWU:
> **Update (2026-08-06):** the impact figures in the original description were
> overstated, and the root cause of the surrounding slowness turned out to be a
> separate bug — a sandbox timeout that abandons a spinning thread (#623). The
> two problems described here are real, independently reproducible, and still
> worth fixing; they are just smaller than first claimed. Corrected attribution
> is in the comment below. The original text is kept intact for the record.

## Description

The backend unit suite spends a meaningful part of every CI run talking to the network — or rather, waiting on network it can't reach. Two independent causes, both in code that runs during ordinary unit tests:

1. **Every Celery dispatch blocks ~19 seconds** retrying Redis. Fully diagnosed below. (Originally described here as "the dominant cost" — see the update above; it is ~82s of a leak-free serial run, second to #623.)
2. **`tests/test_observability.py` leaves a live Sentry client installed** for the remainder of the pytest session.

`backend-test` in `.github/workflows/ci.yaml` declares a `mongo` service but **no redis**, and runs a `["3.11", "3.12"]` matrix — so both jobs pay (1) in full on every push and PR.

For reference, a full serial run on my machine: **807.86s**, 3157 passed / 165 skipped.

---

# 1. Celery dispatches retry an absent Redis result backend (~19s each)

## Evidence

Slowest tests from `pytest -q --durations=40`:

```
59.76s call     tests/test_observability.py::TestInitSentry::test_llm_integrations_are_disabled
40.02s call     tests/test_extraction_tasks.py::TestPerformExtractionTask::test_updates_activity_to_running_then_completed
37.09s call     tests/test_web_fetcher.py::test_static_html_extracts_main_content
31.99s call     tests/test_output_handlers.py::TestSaveResultsToFolder::test_saves_csv_file
20.29s call     tests/test_extraction_tasks.py::TestPerformExtractionTask::test_handles_missing_activity_gracefully
```

`test_updates_activity_to_running_then_completed` mocks the database, mocks `ExtractionEngine`, and asserts on `update_one` call counts. Nothing in it should take 40 seconds. Profiling it:

```
         3142685 function calls (3073510 primitive calls) in 39.436 seconds
   ncalls  tottime  percall  filename:lineno(function)
       76   38.007    0.500   {built-in method time.sleep}

Function                      was called by...
[...]

PR #624, opened 2026-08-06
fix(sandbox): stop runaway sandbox threads instead of abandoning them

### LINKED ISSUE #623: [Bug] Sandbox timeout abandons a spinning thread — pegs a Celery worker core permanently, and is ~87% of backend test runtime
## Description

`execute_sandboxed_code` runs user code in a **daemon thread** and, when that
code exceeds its timeout, returns `{"timed_out": True}` while leaving the thread
running. Python cannot kill a thread, so code like `while True: pass` keeps
spinning for the remaining life of the process.

The timeout *reports* success while the runaway continues. In a Celery worker
that thread survives every subsequent task.

This is also, by a wide margin, the largest cost in the backend test suite:
two tests that exercise the timeout path account for **~87% of a serial run**.

## The code

`backend/app/utils/code_sandbox_runner.py:63-70`:

```python
thread = threading.Thread(target=_run, daemon=True)
thread.start()
thread.join(timeout=timeout)

if thread.is_alive():
    return {"timed_out": True}   # <-- thread is still running, and always will be
```

## Demonstration

Calling it once with a one-second timeout, then measuring the host process:

```
threads before       : 1
io baseline          : 9.3ms for 2000 read+truncate
sandbox returned     : {'timed_out': True}
threads after        : 2
cpu burned in 2s idle: 2.02s      <- a full core, spinning, indefinitely
io after leak        : 37388.8ms for the identical 2000 operations
slowdown factor      : 4027.5x
```

The 4027x is a deliberately GIL-sensitive microbenchmark (tight `read`/`truncate`
syscalls), so treat it as an upper bound rather than a typical figure. The
mechanism is GIL contention: the spinning thread holds the GIL between switch
intervals, and `sys.getswitchinterval()` is `0.005`, so operations that release
and reacquire the GIL pay ~2 intervals each.

For a real-workload number, see the suite measurements below.

## Production impact

Two callers, both reached from Celery workers:

- `backend/app/services/workflow_engine.py:683-686` — workflow Code node,
  `CODE_TIMEOUT_SECONDS = 10`
- `backend/app/services/cross_field_validation.py:326` — validation-rule
  expression, `timeout=5`

`validate_sandbox_code` runs first and blocks imports, `exec`, `eval`, attribute
introspection and so on — but it does **not** block loops. `while True: pass` is
exactly what the existing test uses, and an accidental unbounded loop in a
user-authored validation expression or Code node reaches this path.

Neither `app/celery_app.py` nor `run_celery.sh` sets `worker_max_tasks_per_child`,
so prefork children are long-lived. Leaked threads therefore accumulate: each
timeout permanently costs that worker one core and slows every task it later
runs. `task_time_limit = 3660` does not help, because the task itself completes
normally — it is only the abandoned thread that keeps running.

## Test-suite impact

Full backend suite, serial, same machine, measured back to back:

| configuration | wall | cpu_user |
|---|---|---|
| as shipped | **729.70s** | 735.4s |
| deselecting the 2 timeout tests | **11.39s** | 17.9s |

The two tests are:

- `tests/test_code_execution.py::TestCodeExecutionTimeout::test_busy_loop_times_out_without_hanging`
- `tests/test_workflow_nodes.py::TestCodeExecutionNode::test_timeout`

Deselecting only the first, in an 80-file subset: **208.10s -> 6.65s**.

Both tests are correct and worth keeping — they are testing real behaviour. The
problem is the behaviour they leave behind.

## Suggested fix

The thread has to become interruptible or killable. Two options, with real
trade-offs:

1. **Killable subprocess.** Reinstate a `multiprocessing` process (the module
   still carries `run_sandboxed_code`, described as the "legacy entry point for
   multiprocessing-based execution") and `terminate()`/`kill()` it on timeout.
   Robust against any runaway, including C-level loops. Costs process startup
   per call, and `input_data` must be picklable — worth checking against what
   the Code node actually passes. `fork` in a threaded Celery worker is risky;
   `spawn` is safer but slower.

2. **Interruptible thread.** Install a per-thread trace function that raises
   once the deadline passes. No new process, no pickling, no startup cost, and
   it stops pure-Python loops — which is what both the tests and the realistic
   accident look like. It cannot interrupt a runaway inside a single C call
   (e.g. `sum(range(10**18))`, both names being in `_SAFE_BUILTINS`), so it is a
   narrower guarantee.

I lean toward (2) for a contained fix plus an explicit note about the C-level
gap, but (1) is the only option that is airtight, and the choice depends on how
much you want to spend per Code-node execution.

## Acceptance Criteria

- After a sandbox timeout, `threading.active_count()` returns to its prior value
  and the process burns no CPU while idle.
- Both timeout tests still pass and still assert the timeout is reported.
- A full serial run of the backend suite completes in seconds rather than
  minutes, with no tests deselected.

arhyneRWU:
Fixes #623. Also fixes both items in #615.

## What this fixes

`execute_sandboxed_code` ran user code in a daemon thread and, on timeout,
returned `{"timed_out": True}` while leaving that thread running. Python cannot
kill a thread from outside, so `while True: pass` keeps spinning for the life of
the process, holding the GIL between switch intervals.

**In production** this is permanent per occurrence. `worker_max_tasks_per_child`
isn't set in `app/celery_app.py` or `run_celery.sh`, so prefork children are
long-lived: each sandbox timeout costs that worker one pegged core and slows
every task it later handles. `task_time_limit` doesn't help — the task itself
completes normally; only the abandoned thread keeps running. Reachable from the
workflow Code node (`workflow_engine.py:683`) and validation-rule expressions
(`cross_field_validation.py:326`). `validate_sandbox_code` blocks imports and
`exec`/`eval`, but not loops.

**In CI** it was ~87% of the backend suite runtime.

## The change

Three files, 95 insertions:

- `app/utils/code_sandbox_runner.py` — install a per-thread trace function that
  raises once the deadline passes, so pure-Python loops unwind promptly instead
  of being abandoned.
- `tests/conftest.py` — point Celery at in-memory transport for the unit suite
  (#615). Every dispatch a test reached was blocking ~19s retrying the redis
  **result backend** (20 retries, 1s ceiling), and CI runs no redis service.
  Overriding the broker alone gains nothing: the broker publish already fails
  fast on `ECONNREFUSED`.
- `tests/test_observability.py` — give the real `init_sentry()` a discarding
  transport and tear the global client down afterwards (#615). `sentry_sdk.init()`
  installs a *process-global* client, so the rest of the session was running
  under a live `HttpTransport` aimed at sentry.io with `traces_sample_rate=1.0`.
  The fixture is deliberately module-scoped — applying it suite-wide would
  re-enter `init()` once per test.

## Result

`make backend-test`, serial, same machine:

| | wall | tests |
|---|---|---|
| before | **1425.6s** (23m45s) | 3156 passed, 165 skipped, 58.46% |
| after | **23.7s** | 3156 passed, 165 skipped, 58.44% |

No new dependencies, no tests skipped or deselected, no changes to the Makefile
or CI config. `make backend-ci` and `make backend-static` both pass. Also
verified on Python 3.11 (the other leg of the matrix).

## Trade-offs, stated explicitly

**Tracing costs ~5.4x on sandboxed code.** A 200k-iteration loop goes from 8.1ms
to 43.3ms. Realistic Code-node and validation-expression payloads are orders of
magnitude smaller and sit alongside LLM calls taking seconds, so the absolute
cost is negligible — but it is a real regression and I'd rather flag it than
bury it. Sampling the clock instead of checking every line only recovered ~5%,
so it wasn't worth the extra state.

**A runaway inside a single C call is still uninterruptible.** e.g.
`sum(range(10**18))` — both names are in `_SAFE_BUILTINS` — executes no Python
lines, so the trace function never fires. Only a killable subprocess closes that
gap, at the cost of process startup and requiring picklable input. Both options
are written up in #623; happy to switch if you'd prefer the stronger guarantee.

## What I deliberately did not include

Earlier drafts of this work added `pytest-xdist`, `--capture=sys`, and
`COVERAGE_CORE=sysmon`. All three are dropped. They were compensating for this
bug rather than fixing it, and once the root cause is addressed they earn
essentially nothing while adding a dependency, parallel-execution risk, and the
loss of subprocess output on failing tests. See #622 for that dead end and how
it was ruled out.

### DIFF
--- a/backend/app/utils/code_sandbox_runner.py
+++ b/backend/app/utils/code_sandbox_runner.py
@@ -9,7 +9,9 @@ import json
 import math
 import queue
 import re
+import sys
 import threading
+import time
 from typing import Any
 
 
@@ -41,6 +43,14 @@ _SAFE_BUILTINS = {
 }
 
 
+class _SandboxTimeout(BaseException):
+    """Raised inside the sandbox thread to unwind it once the deadline passes.
+
+    Derives from BaseException so sandboxed ``except Exception`` handlers can't
+    swallow it and keep the thread alive.
+    """
+
+
 def execute_sandboxed_code(code: str, input_data: Any, timeout: int = 10) -> dict[str, Any]:
     """Execute sandboxed code in a daemon thread with a timeout.
 
@@ -48,23 +58,60 @@ def execute_sandboxed_code(code: str, input_data: Any, timeout: int = 10) -> dic
     - ``{"result": <value>}`` on success
     - ``{"error": <message>}`` on runtime error
     - ``{"timed_out": True}`` when the code exceeds the timeout
+
+    The thread is *stopped* on timeout, not abandoned. Python has no way to kill
+    a thread from outside, so we install a per-thread trace function that raises
+    once the deadline passes; because tracing fires per line, any pure-Python
+    loop unwinds promptly.
+
+    Abandoning it is not a survivable option: a runaway like ``while True: pass``
+    would keep spinning for the life of the process, holding the GIL between
+    switch intervals and degrading everything else in it. In a Celery worker
+    that is permanent, and it accumulates one pegged core per timeout.
+
+    Two costs, both measured and both judged acceptable:
+
+    - Tracing makes sandboxed code ~5.4x slower (a 200k-iteration loop goes from
+      8.1ms to 43.3ms). Realistic Code-node and validation-expression payloads
+      are orders of magnitude smaller than that, and sit alongside LLM calls
+      taking seconds, so the absolute cost is negligible. Sampling the clock
+      rather than checking it every line only recovered ~5%, so it isn't worth
+      the extra state.
+    - A runaway *inside a single C call* (e.g. ``sum(range(10**18))`` — both
+      names are in ``_SAFE_BUILTINS``) executes no Python lines, so the trace
+      function never fires and that call still cannot be interrupted. Only a
+      killable subprocess closes that gap, at the cost of process startup and
+      requiring picklable input. Worth revisiting if it shows up in practice.
     """
     result_holder: dict = {}
     local_vars = {"data": input_data, "result": None}
+    deadline = time.monotonic() + timeout
+
+    def _tracer(frame, event, arg):  # type: ignore[no-untyped-def]
+        if time.monotonic() >= deadline:
+            raise _SandboxTimeout
+        return _tracer
 
     def _run() -> None:
+        sys.settrace(_tracer)
         try:
             exec(code, {"__builtins__": _SAFE_BUILTINS}, local_vars)  # noqa: S102  # nosec B102
+        except _SandboxTimeout:
+            result_holder["timed_out"] = True
+            return
         except Exception as exc:
             result_holder["error"] = str(exc)
             return
+        finally:
+            sys.settrace(None)
         result_holder["result"] = local_vars.get("result")
 
     thread = threading.Thread(target=_run, daemon=True)
     thread.start()
-    thread.join(timeout=timeout)
+    # Allow a little slack past the deadline for the tracer to unwind the stack.
+    thread.join(timeout=timeout + 1.0)
 
-    if thread.is_alive():
+    if thread.is_alive() or result_holder.get("timed_out"):
         return {"timed_out": True}
 
     return result_holder if result_holder else {"result": local_vars.get("result")}
--- a/backend/app/utils/code_sandbox_runner.py
+++ b/backend/app/utils/code_sandbox_runner.py
@@ -4,15 +4,17 @@ This module is intentionally kept free of heavy dependencies (no httpx,
 BeautifulSoup, pydantic-ai, MongoDB drivers, etc.) to keep it fast to import.
 """
 
+import ctypes
 import datetime
 import json
+import logging
 import math
 import queue
 import re
-import sys
 import threading
-import time
 from typing import Any
+
+logger = logging.getLogger(__name__)
 
 
 _SAFE_BUILTINS = {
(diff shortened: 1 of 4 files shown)


arhyneRWU:
## Impact on this project's own CI

Not a local benchmark — these are GitHub Actions runs on this repo:

| Backend Tests (py3.12) | wall |
|---|---|
| #621 (current `main` behaviour) | **43m 12s** |
| this PR | **1m 58s** |

**~22x faster.** Same test count, same coverage gate, no tests skipped.

It should also remove the run-to-run variance: on #621 the identical suite took
**15m53s** on py3.11 but **43m12s** on py3.12. GIL contention from the leaked
threads is timing-dependent, so the suite's duration has been close to random.

<sub>Caveat: #621 is a different branch (it carries the extraction-quality-gate
changes), so this isn't a perfectly controlled A/B — but a handful of extra
tests can't account for 41 minutes, and it matches the controlled local
measurement in the description: 1425.6s → 23.7s.</sub>

arhyneRWU:
## Hold — this fix does not work on Python 3.11. Moving to draft.

Correcting my own comment above before anyone acts on it. The 22x figure is real
for py3.12, but I quoted it while **half the supported matrix was still broken**,
and I should have checked before posting it.

**What's wrong.** The timeout works by installing a trace function that raises
once the deadline passes. That relies on `line` events firing repeatedly inside
the runaway loop. On 3.11 they don't:

| sandboxed code | py3.11 | py3.12 |
|---|---|---|
| `while True: pass` (one line) | **1 line event — thread survives** | 201 events — thread stops |
| `while True:` / `    pass` (two lines) | 201 events — thread stops | 201 events — thread stops |

On 3.11 a loop back-edge does not re-emit a `line` event when the line number
doesn't change, so the tracer fires once and never again. 3.12 changed this.
The test in this repo uses the single-line form, so 3.11 is exactly the broken
case.

Measured on 3.11 with this branch applied:

```
sandbox returned     : {'timed_out': True}
threads after        : 2          <- should be 1
cpu burned in 2s idle: 3.01s      <- should be 0.00s
slowdown factor      : 5849.1x
```

That also explains the CI timings, which I had put down to a slow runner:
py3.12 **1m58s**, py3.11 **32m25s** on the same commit.

**Why CI didn't catch it.** The existing tests assert only that
`{"timed_out": True}` is returned — which the *original, unfixed* code also did.
They cannot tell a stopped thread from an abandoned one, so 3.11 passed green
while still leaking. That gap is the more important finding here, and the fix
needs a regression test asserting the observable that actually matters: thread
count returns to baseline and no CPU is burned afterwards.

**Next**, in order:

1. Add that regression test — it should fail on 3.11 against the current commit.
2. Replace the mechanism with a version-independent one. Either
   `frame.f_trace_opcodes = True` (fires per bytecode rather than per line, so
   single-line loops are caught, at higher tracing cost) or a watchdog raising
   via `PyThreadState_SetAsyncExc` (no steady-state overhead, but reaches for a
   CPython C-API call through `ctypes`). Input welcome on which you'd prefer.
3. Re-verify on both 3.11 and 3.12 before this leaves draft.

The diagnosis in #623 is unaffected — the leak, its production impact, and the
attribution all still hold. It is only this particular remedy that is
incomplete.

arhyneRWU:
## Resolved — both Python legs fixed, full numbers

The 3.11 problem flagged above is fixed (async exception instead of line-based
tracing) and CI is green across all 10 checks. Superseding my earlier comment,
which quoted py3.12 only:

| Backend Tests | #621 (without fix) | this PR |
|---|---|---|
| **py3.11** | 15m 53s | **2m 13s** |
| **py3.12** | 43m 12s | **2m 8s** |

The variance is gone too, which is the clearer signal: on #621 the two legs of
the same matrix differed by **27 minutes** on identical code. Here they differ
by **5 seconds**. That spread was the leaked threads — GIL contention is
timing-dependent, so the suite's duration was close to random.

Also worth noting the new mechanism is cheaper than the one it replaces: the
first attempt used a per-line trace function costing ~5.4x on ordinary
sandboxed code. This one costs nothing until a timeout actually fires
(measured 1.0x).

<sub>Caveat unchanged: #621 is a different branch, so this isn't a perfectly
controlled A/B — but it matches the controlled local measurement in the
description (1425.6s → 23.7s), and a few extra tests can't account for 41
minutes.</sub>

ViaJables:
Nice one! It is great having your help!

PR #626, opened 2026-08-10
fix(citations): hedge page numbers estimated from OCR text

arhyneRWU:
> **Ready for review.** Rebased onto `main` now that #604 has merged, so the diff is just this PR's three commits. Backend `3404 passed, 165 skipped`.

Refs #603.

## The defect

Scanned PDFs have no page structure to read. `_interpolate_page_markers` estimates each boundary by spreading the page count evenly across the OCR text, and those markers are shaped identically to the measured ones from `_pymupdf_extract_with_pages`. #604 added `approximate: true` to tell them apart and used it in document chat. Every other consumer still presented the guess as a measurement:

| Surface | Before | After |
|---|---|---|
| KB chat source label | `Source (p. 12)` | `Source (p. ~12)` |
| Workflow citations | `Source · p. 12` | `Source · p. ~12` |
| Extraction source pins | `p. 12` | `p. ~12` |

How wrong the guess gets, measured rather than assumed:

| Corpus | Result |
|---|---|
| 36-page scanned packet | interpolated marker lands on the right page **84%** of the time |
| 68-page proposal (#604) | cited page correct **4.5%** of the time, median error **6 pages** |

## What changed

| File | Change |
|---|---|
| `services/page_locator.py` (new) | `location_meta` writes the metadata, `locator_for_meta` renders it. One module so the key ingest writes and the key renderers read cannot drift — a drift would silently un-hedge every page with nothing failing. |
| `services/document_manager.py` | Both ingest paths write `page_approximate` through `location_meta`, replacing two copies of inline page/sheet branching. |
| `services/chat_service.py`, `workflow_engine.py` | Source labels resolve through `locator_for_meta`. |
| `services/extraction_sources.py` | `page_for_offset` reimplemented on `page_marker_for_offset` so callers see provenance, not just the number. Signature and behaviour unchanged. |
| `frontend/src/utils/pageLocator.ts` + 3 panels | Renders the locator; the three citation types gain optional `page_approximate`. |

**The flag is written only when true.** Measured pages keep exactly today's shape, so stored results stay comparable and nothing needs re-indexing. Citations written before the flag render as measured until reprocessed. That constraint came from a test failure, not foresight — the first version set `page_approximate: False` on every sidecar and an existing shape assertion caught it.

## Measured against live models

Qwen3-VL-30B and 8B, temperature 0, context cleared between runs, synthetic CSU corpus only.

**The 30B ignored the old hedge instruction.** Document is 100% interpolated markers (verified in the database, 3 of 3 flagged):

```
before   ...explicitly stated in the budget summary table on page 1
after    ...explicitly stated in the budget table on page ~1
```

5/5 runs stripped the tilde before; 5/5 keep it after. The 8B hedged correctly both before and after — the failure is model-specific, which is why the instruction has to be robust rather than merely present.

**The old notes were asymmetric.** Measured pages got *"cite when you quote"* (conditional, so a summary answer owes nothing); interpolated pages got an unconditional directive. The trustworthy document carried the weaker instruction. 30B on a digital PDF with measured markers:

| Question | Before | After |
|---|---|---|
| "What is the total amount requested?" | 0/1 | **5/5** |
| "Is the imaging flow cytometer in the MTDC base?" | 0/1 | **5/5** |
| "List the three budget periods and the total for each" | 0/1 | **5/5** |

`.docx`, which has no page markers, claimed no page in 5/5 on both models — the note is correctly omitted rather than inviting invented numbers.

## Evidence

| Claim | Test that proves it |
|---|---|
| Frontend actually hedges | `ChatMessage.citations.test.tsx` renders the real component; removing the prop makes it fail |
| A *new* citation surface can't skip the hedge | `pageLocator.callsites.test.ts` greps every call site and asserts ≥5 found, so it can't pass by matching nothing |
| Persistence chain holds | `test_chunk_page_metadata.py` — verified by reverting the fix and confirming the right tests fail |
| Writer/reader contract holds | `test_page_locator.py` |
| Instruction shape | `test_page_notes.py` (12 tests) |

## Limits of this

| Limit | Detail |
|---|---|
| Partial compliance | The 30B still writes "explicitly", which the new note forbids by name. The tilde survives, so the substantive signal is there — but this is not a clean win. |
| Asymmetric run counts | Baseline was 1 run per question, follow-up 5. A five-run baseline needs the old prompt rebuilt and is worth doing before anyone quotes these as rates. |
| Not byte-deterministic at temp 0 | Five identical requests gave four identical answers and one that swapped "section" → "table" (both correct). Read every row as an observation, not a guarantee. |
| Pre-flag documents | No `approximate` value stored, so they render as measured until reprocessed. |

## Test Plan

- [x] Shared checks pass (`make ci`) — backend `3404 passed, 165 skipped`, coverage 60% against a 50% gate; frontend typecheck / lint / audit / build clean. Frontend half runs in a pinned container; this machine has no local node.
- [x] `make backend-static` — ruff and bandit clean
- [ ] Release check passes if packaging or deployment changed (`make release-check`) — n/a
- [x] Manually tested the affected feature(s) — hedged citations confirmed in a browser against scanned, digital and `.docx` documents on two models
- [x] Updated `CHANGELOG.md`
- [ ] Updated deploy/release docs — n/a


ViaJables:
Heads-up that everything this was stacked on has landed, so it's unblocked whenever you want to pick it up.

Merged today: #604 (page count + `[p. N]` annotation, with the `approximate` flag this PR builds on), #614, #631, #632, and #641. The CHANGELOG entry from #604 also got moved out of the released `v4.10.0` section into `[Unreleased]`, where it belonged.

Two of those are worth knowing about before you rebase:

- **#632** (long-document model routing) touches `chat_service` around the same segment-building path — `build_document_segments` from #604 now feeds the pre-compaction token estimate. Your page-note wording change should still sit cleanly on top, but that's the area to look at. You'd flagged the notice-heading regression as depending on #604/#626; that dependency is half-resolved now.
- **#645** (OCR outage retry, closes #633) changes *when* OCR failures happen at all — a transient outage now retries for ~25 minutes instead of falling through to PyMuPDF after three seconds. That doesn't change what an interpolated marker means, but it does mean fewer documents reach the interpolated path by accident, which is arguably good news for the coverage numbers you were measuring.

No rush, and no need to re-run the model evaluation on our account — the accuracy work in #604 is what merged, and the coverage question this PR addresses is still the open one. Ping me when it's out of draft and I'll review it.

PR #627, opened 2026-08-11
fix(support): declare queue filters once so partial search updates typecheck

arhyneRWU:
## What's broken

`main` fails `make frontend-ci` at the typecheck step with four `TS2322` errors. That blocks CI on every branch cut from it, so no PR can show a green frontend check right now.

```
src/components/layout/TeamsDropdown.tsx(225,17): error TS2322
src/pages/SupportCenter.tsx(107,7):  error TS2322
src/pages/SupportCenter.tsx(231,32): error TS2322
src/pages/SupportCenter.tsx(237,32): error TS2322
```

## Why

The Support Center's URL filters are declared in two places:

- `router.tsx` — `validateSearch` returns all six keys, which TypeScript infers as **required**-but-possibly-undefined
- `SupportCenter.tsx` — the same shape with **optional** keys

Neither is wrong alone, but no caller can satisfy both. Every partial update fails:

```tsx
<Link to="/support" search={{ ticket: undefined }} />
navigate({ search: (prev) => ({ ...(prev as SupportSearch), ticket: uuid }) })
```

TypeScript reads `{ ticket: undefined }` as *missing* `status`, `priority`, `classification`, `tag` and `q` — even though all five would be `undefined`.

## The change

Moved the filter types next to the other Support types in `types/support.ts` and annotated `validateSearch` with the shared type, so the route and the page agree by construction instead of by hand.

Optional is the honest shape: `validateSearch` already returns `undefined` for every absent or unrecognized value, so an omitted key and a key set to `undefined` mean the same thing, and a caller should be free to pass either.

**Types only — no behaviour change.** No runtime code was touched.

## Verification

| Check | Result |
|---|---|
| `tsc -b` on unmodified `main` | **4 errors** |
| `tsc -b` with this change | **0 errors** |
| `eslint` on all changed files | clean |
| `vitest run` | 495 passed, 64 files |

Found while investigating a CI failure on an unrelated PR — the failing files aren't touched by that branch.

### DIFF
--- a/frontend/src/pages/SupportCenter.tsx
+++ b/frontend/src/pages/SupportCenter.tsx
@@ -15,12 +15,10 @@ import {
 } from '../api/feedback'
 import type {
   SupportTicket, SupportTicketSummary, SupportAttachment,
+  StatusFilter, PriorityFilter, ClassificationFilter, SupportSearch,
 } from '../types/support'
 
 type View = 'list' | 'new' | 'chat' | 'whats_working'
-type StatusFilter = 'all' | 'open' | 'in_progress' | 'closed'
-type PriorityFilter = 'all' | 'low' | 'normal' | 'high'
-type ClassificationFilter = 'all' | 'bug' | 'enhancement' | 'feature_request'
 
 const MAX_BYTES = 10 * 1024 * 1024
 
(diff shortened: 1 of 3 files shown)


arhyneRWU:
Superseded by d0928371, which landed the same fix independently — the `/support`
search shape moved into `types/support.ts` with optional keys and
`validateSearch` annotated with it, which is what this PR did. Nothing here is
missing from `main`. Closing.

PR #637, opened 2026-08-13
fix(authz): verification reviewers could not open what they were sent to review

ViaJables:
## The bug

A reviewer received the "New submission awaiting review" email for a workflow submitted by a user on another team. Clicking through to the item gave **"Workflow not found."** Reported from the field on 2026-08-13.

## Root cause

Two different scopes that never lined up:

- **Reviewer eligibility is global.** `_notify_examiners` fans out to `{"$or": [{"is_admin": True}, {"is_examiner": True}]}` org-wide (`verification_service.py:1590-1592`), keyed off a single flag, `User.is_examiner` (`models/user.py:13`).
- **Item access is team-scoped.** `can_view_workflow` (`access_control.py:461-473`) is owner-or-team-member, with no notion of a verification reviewer.

So the queue had been mailing people links to items they structurally could not open. This was not an edge case — it hit *every* examiner outside the submitter's team, on *every* submission. The library-backed fallback doesn't help, because a submission still awaiting review isn't in the verified library yet.

Knowledge bases and search sets had the same gap: their carve-outs key on `verified`, which is false for a pending submission.

## The fix

New helper `has_open_verification_review_access(item_kind, object_id, user)` — grants access when the user is an examiner or admin **and** an open `VerificationRequest` exists for the item. Open means `submitted` / `in_review` / `returned`; approved items are reachable through the verified library anyway, and rejected ones correctly revert to private.

Wired into all five getters for the three kinds the queue supports, each guarded by `not manage` so it grants **view only** — reviewers read and test-run a submission, never edit it:

- `get_authorized_workflow`
- `get_authorized_knowledge_base` / `..._by_id`
- `get_authorized_search_set` / `..._by_id`

## No frontend change

`can_manage` is computed separately (`workflow_service.py:122`) and stays `false` for a non-team examiner, which the editor already honors as view-only (`WorkflowEditorPanel.tsx:605,1141`). Reviewers get a read-only, runnable view — exactly the review workflow.

## Testing

- 11 new tests in `test_access_control.py`: the cross-team regression, that the carve-out never grants manage, non-examiners short-circuiting before touching the DB, admin parity, no-open-request denial, and a guard on the query predicate asserting `approved`/`rejected`/`draft` are excluded.
- 96 pass in that file; 336 across the workflow / verification / KB / search-set / authz suites.
- Full backend unit suite: 3253 passed. One unrelated failure, `test_code_sandbox_timeout_cleanup.py::test_single_line_runaway_leaves_no_live_thread` — sandbox thread timing, untouched here, passes in isolation and flakes under full-suite load.
- Ruff and Bandit clean on the changed file.

## Reviewer note

`is_admin` is included alongside `is_examiner` because admins receive the same notification and hit the same wall. That does mean admins gain read access on a path that otherwise passes `allow_admin=False` — scoped to open requests only. Easy to restrict to examiners if you'd rather.

## Follow-up not in this PR

The notification email links to `{frontend_url}/verification` (`email_service.py:426`), not to the specific request, so reviewers land on the queue and have to find the row. The approval flow already does this properly with `/reviews/{uuid}`.


ViaJables:
Belatedly answering the reviewer note in this PR, since it merged without a response and it's a question about an authorization widening — those shouldn't land on silence.

**On including `is_admin` alongside `is_examiner`: keep it as you built it.** `_notify_examiners` fans out to `{"$or": [{"is_admin": True}, {"is_examiner": True}]}`, so admins receive the same mail and hit the same wall. Granting access to the notified set and not to everyone in it would leave exactly the bug this PR fixes, just for a smaller group.

The scoping is what makes it fine: view-only (`not manage`), and only while an open `VerificationRequest` exists for that item — approved items are reachable through the verified library anyway, rejected ones revert to private. That's a much narrower grant than a general `allow_admin=True`, and it's tied to an action the admin was explicitly asked to perform.

Also confirming the follow-up you flagged is real and now tracked: the notification email links to `{frontend_url}/verification` rather than the specific request, so reviewers land on the queue and hunt for their row while the approval flow already does this properly with `/reviews/{uuid}`. Filed separately.

Issue #629, opened 2026-08-12
[Feature] Per-model token calibration — designed, not built, and I'd rather ask before building it


arhyneRWU:
## Problem

Token budgets for a model whose vocabulary cannot be obtained — hosted APIs — are computed with a **guessed** safety margin. The divergence I measured is content-dependent, not model-dependent: a self-hosted Qwen tokenizes digits individually where `cl100k_base` groups them, so prose comes out identical under both while a dense budget table came out 45% higher than estimated. Research-administration documents sit at the digit-dense end, so the error lands hardest on this product's core content.

A single constant cannot cover both: one tuned for currency tables wastes a third of the window on prose, and one tuned for prose hard-fails budget documents. Counting with the model's own vocabulary fixes this wherever a `tokenizer.json` is on disk, but hosted providers do not publish one — so for those models, **measurement is the only route to a real number**.

## Proposed Solution

Replace the guessed margin with a measured, stored, per-model value obtained once from an admin-initiated calibration, rather than inferred from live traffic. Fixtures in three content families (prose, mixed, numeric), each sent at two payload sizes so a multiplicative rate and a fixed per-request overhead can be separated — one measurement cannot distinguish them.

The stored result takes the worst rate across families rather than the mean, because the costs are asymmetric: guessing high routes a request early, guessing low hard-fails it.

**I would rather agree this before writing any of it.** It is a new service module, an admin endpoint, a ladder change in `context_budget`, and UI.

## Alternatives Considered

- **Provider-native token-counting endpoints** for hosted models, which several providers offer. The design deliberately leaves this open, and it may be the better answer for that population — direction welcome.
- **Inferring calibration from live traffic** instead of an explicit action. Rejected: it makes budgets drift silently and is hard to reason about when something goes wrong.
- **Keep the guessed constant.** Workable but knowingly wrong for numeric content, which is this product's core material.
- **Close this and do nothing.** Genuinely fine — the two built parts stand on their own.

## Additional Context

<details>
<summary><b>Full detail — where the work sits, the design, and where the fixtures would come from</b></summary>

**Heads-up first, so this issue isn't confusing: none of the work described here
is in the project yet.** I hit a token under-count on a self-hosted deployment,
and the fix came out in three parts. Two are built and tested on a branch, and I
will open them as PRs in the next few days once my currently open PRs are out of
your way:

1. **Exact tokenization** — count with the model's own `tokenizer.json` where one
[...]

PR #648, opened 2026-08-17
fix(chat): count tokens with the model's own vocabulary

arhyneRWU:
Splits the exact-tokenizer half out of #629, which asks for agreement before building per-model calibration. Nothing here needs that agreement: no new service, no admin endpoint, no UI. The calibration design question stays open on #629.

## The bug

`_encoding_for()` returns tiktoken's `cl100k_base` for every model that is not GPT-4o/4.1/o-series. That is OpenAI's tokenizer, used as a proxy for Claude, Llama, Mistral and every self-hosted model. On this deployment it covers 100% of configured models.

It reads **low**, which is the direction that hard-fails.

## Why it matters more since #632

Routing compares that same estimate against the budget. An under-count means it sees headroom that isn't there, so it declines to route — and the request is then neither compacted nor routed, and the model rejects it outright:

```
plan:  input_budget 24,576   total_input_tokens 23,592   -> "fits, 984 spare"
then:  {"kind":"error","content":"This conversation is too large for the
        selected model. Remove some documents or switch to a larger model."}
```

A 36-page proposal produced no answer while a 262k model sat configured and unused. The feature built to prevent that was disabled by the arithmetic feeding it.

## The divergence is content-dependent, not model-dependent

This is the part that rules out simply picking a better constant. Measured against the models' own reported `prompt_tokens`:

| Content | model / tiktoken |
|---|---|
| flowing prose | 1.000 |
| project description | 1.019 |
| real budget justification | 1.171 |
| synthetic dense currency table | 1.455 |

A self-hosted Qwen tokenizes digits individually where `cl100k_base` groups them, so prose comes out identical while a dense budget table runs ~45% higher. Research-administration documents sit at the digit-dense end, so the error lands hardest on this product's core content.

No single constant works: one tuned for tables wastes a third of the window on prose, one tuned for prose hard-fails budget documents.

## What this does

**Counts exactly where it can.** When a `tokenizer.json` is on disk, tokens are counted with the model's own vocabulary and **no margin is applied at all** — there is nothing left to guess. Vocabulary only: never the weights, never the GPU. Roughly 1.5 ms for a 36-page proposal against 0.3 ms for the tiktoken call it replaces, and the loaded tokenizer is cached.

Discovery uses the standard HuggingFace cache layout and takes the newest snapshot, so a re-pulled model is picked up without config changes. Overridable per model (`tokenizer_path`, `tokenizer_cache_root`) and per deployment (`TOKENIZER_CACHE_ROOT`).

**Falls back honestly where it cannot.** Hosted models have no local vocabulary, so they keep estimate-plus-margin, with `DEFAULT_TOKEN_SAFETY_MARGIN = 1.20`, a per-model `token_safety_margin` override, and a configured value below 1.0 refused — that would re-create this bug by configuration.

The fallback now says so, once per process:

> token counts for 'Qwen-30b' are estimated, not exact: no local vocabulary and no stored calibration. Budgets use a default margin of 1.20, which is known to under-count numeric and tabular content.

That warning exists for the alias case: a model registered as `Qwen-30b` rather than its full name resolves no vocabulary and silently drops to the guess. The guess is a reasonable last resort; not saying so is not.

**Makes request scaffolding explicit.** `REQUEST_SCAFFOLD_TOKENS = 512` covers the chat template and agent preamble, which appear in no string we count. Measured at a flat 13 tokens against a bare server and 37 end-to-end, identical on a 25,000-token request and a 50,000-token one — a fixed cost, not a multiplier. 512 is deliberately far above that: 2% of a 32k budget, cheap insurance against a framework version adding more. This is what the old margin was incidentally covering, and making it explicit is what lets the margin drop to nothing for models we can tokenize.

**Corrects the workflow pre-flight.** `find_oversize_documents` reads stored `token_count` values that are raw tiktoken figures, so the margin is applied at read time rather than trusted from storage. That also corrects documents ingested before any of this existed, with no backfill.

## Known limitation, stated plainly

**The 1.20 default is a guess and is known to be insufficient for the worst case** — the currency table above measured 1.455. It governs only models where exactness is unavailable, for which no measurements exist on this deployment. A deployment that has measured its own should set `token_safety_margin` per model.

Replacing that guess with a measurement is exactly what #629 proposes, and why that issue should stay open.

## Test plan

- [x] `3481 passed, 165 skipped`
- [x] `ruff check app/` clean
- [x] New: `test_exact_tokenizer.py`, `test_token_safety_margin.py`, plus cases in `test_context_budget.py`
- [x] `tokenizers` declared explicitly rather than relied on transitively

One existing test needed adjusting: `test_find_oversize_documents_honors_reserve_override` asserted a 100,000-token document fits a ~119,000-token budget. With the margin applied that is no longer true — correctly, since 100k raw is ~120k real. The margin is pinned to 1.0 there so the test measures the reserve and nothing else; the interaction is covered directly in `test_token_safety_margin.py`.

Refs #629

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,19 @@
 # Optional. Custom LLM endpoint (OpenAI-compatible). Leave empty to use OpenAI directly.
 INSIGHT_ENDPOINT=
 
+# Optional. HOST path to your HuggingFace cache, mounted read-only into the api
+# and celery containers. Only useful if you self-host models (vLLM, TGI, Ollama
+# with an HF cache) — set it to your HF_HOME, the directory containing `hub/`.
+#
+# Token counting reads each model's tokenizer.json from here so budgets are
+# exact instead of estimated. It reads the vocabulary only: no weights, no GPU,
+# no downloads.
+#
+# Leave unset and nothing breaks — budgets fall back to a tiktoken estimate plus
+# a safety margin, and the log says so once per model. Accuracy suffers most on
+# numeric and tabular documents, where tokenizers disagree most.
+HF_CACHE_PATH=
+
 # --- Infrastructure ---
 redis_host=localhost
 
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -106,3 +106,13 @@ TELEMETRY_CONTACT_EMAIL=
 # /api/telemetry/heartbeat ingest route and the admin Telemetry dashboard.
 # Leave false on every normal deployment — both stay completely hidden.
 TELEMETRY_COLLECTOR_ENABLED=false
+
+# --- Token counting ---
+# Where THIS process looks for model vocabularies, in the standard HuggingFace
+# cache layout (<root>/hub/models--Org--Name/snapshots/<rev>/). Set it to your
+# HF_HOME when running outside Docker; under Docker compose the cache is mounted
+# at /hf-cache and the host path is set by HF_CACHE_PATH in the root .env.
+#
+# Optional. When no vocabulary is found the budget falls back to a tiktoken
+# estimate plus a safety margin and logs that it did, once per model.
+TOKENIZER_CACHE_ROOT=/hf-cache
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -43,6 +43,17 @@ class Settings(BaseSettings):
     chromadb_host: str = ""
     max_upload_size_mb: int = 500
 
+    # Where this process looks for model vocabularies, in the standard
+    # HuggingFace cache layout (`<root>/hub/models--Org--Name/snapshots/<rev>/`).
+    # Set it to your HF_HOME. Token counting reads `tokenizer.json` from here —
+    # the vocabulary only, never the weights and never the GPU.
+    #
+    # Optional. When nothing is found the budget falls back to estimating with
+    # tiktoken plus a safety margin, and says so in the log once per model. That
+    # is the pre-existing behaviour, so leaving this unset costs accuracy for
+    # self-hosted models but breaks nothing.
+    tokenizer_cache_root: str = "/hf-cache"
+
     # Observability
     sentry_dsn: str = ""
     log_format: str = "json"  # "json" for structured logging, "text" for human-readable
--- a/backend/app/services/chat_service.py
+++ b/backend/app/services/chat_service.py
@@ -583,6 +583,10 @@ async def chat_stream(
     # Measured before any trimming: both routing and the dialog's suggestion
     # ask "could another model have held what the user actually sent?", and the
     # post-compaction total cannot answer that — it always fits by definition.
+    # `model_config` carries this model's token safety margin, so the estimate
+    # is an upper bound rather than an optimistic one. Routing decides off this
+    # number: an estimate that reads low makes the router see headroom that is
+    # not there and decline to move a request that does not fit.
     requested_input_tokens = estimate_input_tokens(
         model_name=model_name,
         system_prompt=system_prompt or "",
@@ -590,6 +594,7 @@ async def chat_stream(
         history=previous_messages,
         documents=doc_segments,
         attachments=attachment_segments,
+        model_config=model_config,
     )
 
     routing = RoutingDecision(model_name, False, "")
--- a/backend/app/services/context_budget.py
+++ b/backend/app/services/context_budget.py
@@ -244,23 +244,9 @@ def token_safety_margin(
     below 1.0 is refused — that re-creates the original bug by configuration,
     leaving the planner optimistic in the direction that hard-fails.
     """
-    if model_config:
-        raw = model_config.get("token_safety_margin")
-        if isinstance(raw, bool):
-            raw = None  # bool is an int subclass; never a meaningful margin
-        try:
-            value = float(raw) if raw not in (None, "") else 0.0
-        except (TypeError, ValueError):
-            value = 0.0
-        if value >= 1.0:
-            return value
-        if value:
-            logger.warning(
-                "ignoring token_safety_margin=%r for %s: a margin below 1.0 "
-                "would under-count the request",
-                raw,
-                model_name,
-            )
+    configured = _configured_margin(model_name, model_config)
+    if configured is not None:
+        return configured
 
     # An exact count needs no correction. Inflating it would only route early.
     if resolve_exact_tokenizer(model_name, model_config) is not None:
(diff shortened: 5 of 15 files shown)


arhyneRWU:
## Validated against a second vocabulary — and it refines the explanation in the description

Pulled `openai/gpt-oss-20b` into the model cache to test the discovery path against a different publisher, and the result changes how the divergence should be described.

### Discovery works unchanged

| Model | vocabulary | path taken |
|---|---|---|
| `openai/gpt-oss-20b` | found | exact |
| `Qwen/Qwen3-VL-8B-Instruct` | found | exact |
| `anthropic/claude-sonnet-4-6` | none | fallback + margin |

No configuration needed — the HF cache layout resolved a second publisher's `models--Org--Name` slug first time.

### The divergence is not "non-OpenAI", it is digit handling

Same two content samples, counted three ways:

| content | chars | cl100k | Qwen3-VL-8B | gpt-oss-20b |
|---|---|---|---|---|
| prose | 12,000 | 2,299 | 2,341 (**1.018**) | 2,287 (**0.995**) |
| budget table | 4,392 | 1,122 | 1,608 (**1.433**) | 1,128 (**1.005**) |

gpt-oss tracks `cl100k_base` within half a percent on *both* families. The mechanism is visible directly:

```
"$1,169,898.51 and $182,636.20"

Qwen3-VL-8B   25 tokens  ['$','1',',','1','6','9',',','8','9','8','.','5', …]
gpt-oss-20b   15 tokens  ['$','1',',','169',',','898','.','51',' and',' $', …]
cl100k        15 tokens
```

Qwen splits every digit; gpt-oss groups them, because `o200k_harmony` is an OpenAI-family vocabulary (200,019 tokens) and inherits that behaviour. Qwen's is 151,669 and does not.

### What this means for the PR

**It strengthens the fix and weakens one sentence of the rationale.** The description says tiktoken "for anything else is a proxy that reads low." That is too broad — it reads low for *Qwen-like* vocabularies and is accurate to within 0.5% for OpenAI-family ones.

The useful conclusion is stronger than the original claim: **you cannot predict the divergence from the vendor.** Two self-hosted models, both non-OpenAI by provenance, sit at 1.433 and 1.005 on identical input. No heuristic based on model family, publisher, or "is it hosted" would get this right. Only tokenizing does.

That is also the case for keeping #629 open rather than picking a better constant. A hosted model could sit at either end, and 1.20 would be wasteful for one and dangerous for the other — which is exactly why the fallback logs that it is a guess.

### Caveat on this measurement

The `prose` sample is 12,000 characters of one project description and the `table` sample is a budget table repeated to a usable length, so these are two content families from one corpus, not a survey. The digit-splitting mechanism is directly observable rather than inferred, but the specific ratios are corpus-bound and should be read as illustrative.

ViaJables (commented):
Reviewed the code rather than the narrative, and the premise holds: `_encoding_for` on `main` does return `cl100k_base` for everything that isn't GPT-4o/4.1/o-series, so every self-hosted and Claude model has been budgeted with OpenAI's tokenizer. Splitting this out of #629 was the right call — it stands on its own and needs no decision from that issue.

The design I want to call out as correct, because it would have been easy to get wrong: separating `count_raw_tokens` from `count_tokens`, applying the margin once per aggregate rather than per fragment, and keeping `stored_count_margin` distinct from `token_safety_margin`. That last distinction is subtle and the docstring explains exactly why it exists — a stored `token_count` is a tiktoken figure no matter what we could compute now. Reusing the other function there would have silently under-warned for precisely the models exact tokenization serves.

## One thing to fix before merge

**`find_context_overflow` doesn't get the margin, and it's the one on the workflow path.**

There are two readers of stored `token_count`:

- `find_oversize_documents` → `chat_service.py:674` — corrected by this PR ✅
- `find_context_overflow` → `workflow_tasks.py:825` — **not touched** ❌

`find_context_overflow` landed in #641 while this branch was in flight, so this is almost certainly a base-timing artifact rather than an oversight. But the effect after merge is that the chat path corrects the tiktoken divergence and the workflow path still under-counts by up to the ~45% you measured on digit-dense content — in the same direction that hard-fails.

That's a pointed miss, because `find_context_overflow` exists specifically to stop a package of individually-modest documents being rejected mid-run by the gateway. Grant budget workbooks are both the documents most likely to trip the combined check and the content where tiktoken diverges most.

The call site already resolves and passes `model_config`, so it should be the same shape you used in `find_oversize_documents` — apply `stored_count_margin` to each `token_count` before the per-document and summed comparisons. Worth a test that a group passing raw and failing margined is flagged, mirroring `test_find_oversize_documents_corrects_a_stored_count_for_an_exact_model`.

## Two smaller notes

**`_find_vocabulary` is `lru_cache`d, so "a re-pulled model is picked up without config changes" is only true across a restart.** The cache key is `(model_name, cache_root)`, and `_load_tokenizer` separately caches the `Tokenizer` by path — so after a re-pull a long-lived worker keeps using the old snapshot's vocabulary until the process restarts. Harmless in practice (vocabularies rarely change between snapshots of the same model), but the comment currently promises more than the code does. Either say "on restart" or drop the caching on the discovery half.

**Memory, worth a sentence rather than a change.** `_load_tokenizer` caches up to 32 tokenizers, each a few MB, per process — multiplied across Celery prefork children. Almost certainly fine given how few models a deployment registers, but 32 is generous for the actual ceiling and it'd be good to know you'd thought about it.

## Checks I ran

- Scanned the branch's `uv.lock` with the Trivy tooling that just landed in #647: **29 HIGH, unchanged from `main`.** Declaring `tokenizers` explicitly adds no supply-chain surface, since it was already there transitively. Good call making it explicit rather than depending on that.
- Verified `_is_openai_model` prefix-matches the o-series rather than substring-matching. `"o1" in name` against a registry of self-hosted names would have been a genuine footgun and you avoided it.
- The `REQUEST_SCAFFOLD_TOKENS = 512` reasoning holds up: measuring the overhead as flat across a 2x payload range is what justifies a constant rather than a multiplier, and 512 against a measured 37 is the right direction to be wrong in.

## On the known limitation

Stating that 1.20 is a guess and insufficient for the 1.455 worst case, in the PR body *and* in the constant's comment, is the right way to ship this. It governs only models where exactness is unavailable, and pretending otherwise would be how the next person gets caught. That's also the strongest argument for keeping #629 open, which is where I left it.

Fix the `find_context_overflow` gap and I'm happy to merge this.

ViaJables:
Pushed the `find_context_overflow` fix to this branch rather than leaving you a chore — it's your PR, so revert or rework it freely if you'd rather do it differently.

Same shape and same source as your `find_oversize_documents` change: `stored_count_margin` applied per document before both the individual and summed comparisons. The call site in `workflow_tasks.py` already resolves `model_config`, so nothing else had to move.

Two of your existing fixtures asserted raw totals and would have failed:

- `test_find_context_overflow_flags_combined_package` (92,119)
- `test_workflow_tasks.py::TestPreflightContextOverflow::test_combined_overflow_fails_before_the_engine_runs` (92,119)

Both now pin `token_safety_margin: 1.0`, so they keep measuring the thing they were written to measure — combination detection and pre-flight wiring — rather than becoming assertions about the margin. That's the precedent you set with `test_find_oversize_documents_honors_reserve_override`, so it seemed better to follow it than to update the numbers and make them margin-dependent.

New coverage in `test_context_budget.py`: a two-document package at 50,000 raw against a 56,320-token budget, which fits raw and doesn't at the 1.20 default — plus the OpenAI case staying uninflated. **Verified by sabotage**: forcing `margin = 1.0` inside the function makes the new test fail and the rest pass, so it's pinning the behaviour and not the arithmetic.

`3483 passed, 165 skipped`, `ruff check app/` clean.

The other two notes from my review are still open and both are yours to judge — the `_find_vocabulary` `lru_cache` making "picked up without config changes" true only across a restart, and whether 32 cached tokenizers per prefork child is the ceiling you want. Neither blocks merge as far as I'm concerned.

PR #654, opened 2026-08-18
fix(kb): stop chrome-only pages and comma-split labels from poisoning KB validation

ViaJables:
From a support ticket: the OMB Uniform Guidance (2 CFR 200) catalog KB was "not reliably using its current source content" — a 150-question validation run scored **73.5** with retrieval precision **0.54**, and 16 auto-generated questions were about the Request Access / "Lock" pages rather than the regulation.

**The source content is not stale.** The bundled text carries the 2024 revision values (`$10,000` equipment threshold, `15%` de minimis; `$5,000` appears zero times), and 811 chunks against ~794 predicted from 634,837 bytes confirms the full corpus is indexed. Three separate defects produced the reported symptoms.

### Chrome-only pages were ingested as content

A grants.gov URL in the seed answers HTTP 200 with a page whose body is only site furniture — the .gov banner, nav, a session-timeout dialog — because its content is client-side rendered. It is not empty and not a bot challenge, so the empty-fetch guard and `looks_like_bot_challenge` both passed it, and ~1 kB of padlock boilerplate was embedded as a source titled **"Lock"**. Retrieval served it, and question generation wrote 16 test questions about padlock icons and session expiry. Those scored **0.350** against a KB average of **0.766** — a 4.4-point drag on the run.

`looks_like_boilerplate_only` rejects a short page carrying two or more chrome markers, on the same conservative shape as the bot-challenge gate (a real page carrying the .gov banner passes on length). It runs before embedding, so a rejected page costs no vectors. `repair_gov_kb_sources` counts such a page as broken so already-poisoned sources are found by the existing sweep; its refetch path deletes chunks before re-ingesting, so the junk is purged even where the URL still cannot be read. The dead URL is dropped from the `uniform_guidance` seed.

### Source labels were split on commas

`Subpart D-iii — Monitoring, Reporting, Remedies & Closeout | §§200.328–200.346` is **one** source. Splitting it on commas produced three labels that name nothing, which both failed to match and tripled the denominator of `hits / len(expected_labels)`. Quotes now protect their contents, and a cell containing semicolons treats those as its only separator. A plain comma list still splits exactly as before, so every file that imported correctly still does.

### Unmatched labels were silent

A label naming no source can never be credited, so every question carrying one scores 0 precision on every run — which is how this set came to be validated against a taxonomy (`Subpart E-i — Cost Principles | §§200.400–200.419`) that named nothing in the KB it was scoring. **57 of 150 questions were at 0 by construction.** Precision on questions with resolvable labels measures **0.961**, not 0.54.

The import response and modal now report unmatched labels with their question counts, using the same substring rule the validation run scores with. The check runs after the questions are written and is wrapped, so a failure to compute a warning cannot turn a successful import into an error.

### Impact

Recomputing the run's own formula (judge 40% + retrieval 25% + health 20% + coverage 15%, which reproduces the reported 73.5 to within 0.01), removing only the 16 off-topic questions and the two dead sources and **changing no answer quality at all**:

| Component | Reported | Corrected |
|---|---:|---:|
| Judge | 72.2 | 76.6 |
| Retrieval precision | 54.0 | 96.1 |
| Source health | 88.9 | 100.0 |
| Chunk coverage | 88.9 | 100.0 |
| **Run score** | **73.5** | **≈ 89.7** |

### Not covered here

Existing deployments keep the bad source until it is removed in the UI — seeding adds sources and never removes them. Deleting a source also does not delete the test queries generated from it. The remaining answer-quality gap (16% of clean answers say "not in context" for text that is indexed) is a retrieval-tuning problem: this KB runs stock defaults, `k=8` with reranking `off`, and has never been through the optimizer.

### Testing

- `make backend-ci` — 3448 passed. One unrelated failure, `test_optimizer_inbox.py::TestListInbox::test_view_only_access_shows_row_without_manage_rights`, comes from in-flight work on `optimizer_inbox.py` in the working tree and is not touched by this branch.
- `make backend-static` — clean.
- `make frontend-ci` — 518 passed, build clean.
- New: 6 boilerplate-detection tests, 7 label-separator tests, 2 import-route tests (unmatched-label reporting, and that a failed check cannot fail the import).


ViaJables:
Reviewed this alongside the rest of today's queue. The diagnosis is excellent — three distinct defects correctly separated, and the recomputed run score is a genuinely useful way to show the impact. But two of the fixes don't do what they're meant to, and I'd rather hand these back with reproductions than guess at what you intended.

Everything below I reproduced on `724baefc`.

---

## 1. The quote escape hatch cannot fire for CSV

`csv.reader` consumes the quoting, so `_split_source_labels` never sees a quote character:

```python
csv_line = 'What is the rule?,"Subpart D-ii — Procurement, Property & Subawards"\n'
row = next(csv.reader(io.StringIO(csv_line)))

row[1]                          # 'Subpart D-ii — Procurement, Property & Subawards'   <- no quotes
_split_source_labels(row[1])    # ['Subpart D-ii — Procurement', 'Property & Subawards']
```

That is exactly the two-bogus-labels outcome the PR sets out to prevent. Only the semicolon path works:

```python
_split_source_labels("Subpart D-ii — Procurement, Property & Subawards; Subpart E")
# ['Subpart D-ii — Procurement, Property & Subawards', 'Subpart E']
```

For a quote to survive `csv.reader`, an author would have to write `"""Subpart D-ii — …"""`, and an XLSX author would have to type literal `"` characters into the cell.

The seven new unit tests all pass because they call `_split_source_labels` directly with raw strings, bypassing the CSV layer. Nothing covers a quoted cell end-to-end through `parse_test_query_import` — that's the test I'd add first, since it fails today.

**Knock-on:** the new template row in `ImportTestQueriesModal.tsx:15` demonstrates the broken path, and its own note says *"Source name contains commas, so the cell is quoted"*. Anyone who downloads the template and edits it inherits two labels naming nothing, which score 0 retrieval precision on every future run. The help text at line 100 ("or quote any source name that itself contains a comma") is wrong for CSV too.

Python's `csv` doesn't expose per-field quoting, so making quotes work is a real design decision rather than a patch — which is why I'm leaving it to you. Making semicolons *the* documented separator and fixing the template would be the smaller, honest version.

## 2. `looks_like_boilerplate_only` rejects real short .gov pages

Every marker except the session-timeout pair is part of the standard USWDS banner that appears on **every** .gov page, so `_MIN_CHROME_MARKERS = 2` is met by the banner alone. And `"locked padlock icon"` is a substring of `"a lock ( lock locked padlock icon )"`, so a *single* lock sentence already scores 2 hits — the "two distinct markers" invariant the comment describes isn't what the code implements.

```python
# standard .gov banner + ~1200 chars of real policy text, 1545 chars total
looks_like_boilerplate_only(page)   # True   <- real content, rejected

looks_like_boilerplate_only(
    "A lock ( Lock Locked padlock icon ) means you are secure. "
    "The de minimis rate is 15 percent."
)                                   # True   <- 92 chars, one lock sentence
```

This isn't the same shape as `looks_like_bot_challenge`, whose markers ("robot or human") appear *only* on interstitials — length plus one marker is safe there because the marker itself is decisive. Here the markers are on every page in the corpus.

Concretely: a user adds a short NIH or grants policy notice under 4,000 chars, and `_ingest_url_source` sets `status="error"` with "content is JavaScript-rendered and did not load". The real content is never indexed.

Deduping the nested lock phrases and requiring markers from distinct *families* would fix the double-count. Beyond that, I think it needs a positive signal that the page is chrome-only — chrome text as a fraction of total text, say — rather than an absolute length gate, since "short" and "chrome-only" aren't the same property.

## 3. That false positive is destructive when the repair sweep runs

`repair_gov_kb_sources.py:123` treats a source whose *stored* content trips the gate as broken. On the default refetch path it calls `delete_kb_source` **first**, then `_ingest_url_source` — which rejects the page again via the same gate.

So a short-but-real .gov source that is currently `ready` with good chunks loses those chunks and ends up `error`. The delete-before-refetch is unconditional, so there's no recovery within the run. Refetching before deleting, and only deleting once the refetch has produced something, would make the sweep safe to run twice.

## 4. `_unmatched_source_labels` can't see document-backed sources

`knowledge.py:1280` resolves a name as `(s.custom_name or s.url_title or s.url or "")`. For `source_type == "document"` both `url_title` and `url` are `None`, so the name is `""` and the source is dropped by the `names = [n for n in names if n]` filter.

But `knowledge_base_tasks.py:137` writes `source_name=doc.get("title", "")` into ChromaDB, and that's what the validation run matches against at `kb_validation_service.py:1380`. So the docstring's claim that this uses "the same substring rule the validation run scores with" doesn't hold for document KBs.

Result: a KB built from uploaded PDFs, with a perfectly labelled test set, warns that *every* label matches no source and invites the user to clear labels that were correct. Mirroring `get_kb_manifest`'s resolution (including `resolve_document_titles`) would fix it.

Smaller, same area: `get_kb_sources` returns rows of any status, so a label naming an `error` or `pending` source is reported as matched even though it has no chunks and will score 0.

## 5. Two smaller ones

- **Mixed separators silently merge labels.** The old code did `.replace(";", ",")`, so mixing was the designed behaviour. Now `Subpart A, Subpart B; Subpart C` goes from three labels to two. The PR body's "every file that imported correctly still does" holds for pure-comma cells only.
- **Quote characters are stripped wherever they appear.** `_split_source_labels('Section 2 "Definitions"')` returns `['Section 2 Definitions']`, which then no longer substring-matches the stored source name. An unbalanced quote also swallows every remaining separator in the cell.

## 6. A stray test class from another branch

`backend/tests/test_knowledge_routes.py:2372` has `class TestTestQueryBulkDelete` containing a `_recording_find` helper and **zero test methods** — and there's no `bulk-delete` endpoint on this branch. It belongs to `feat/kb-test-query-bulk-delete` (#653). Its helper also references `SimpleNamespace`, which isn't imported at module scope here, so it would `NameError` if called. Worth dropping — it'll conflict when #653 lands.

## Also

`VERSION must bump when seeds change` is red: this touches `backend/seeds/knowledge_bases/uniform_guidance.json`, so `backend/seeds/VERSION` needs to go to `1.3.2` or existing installs never see the seed change.

---

To be clear about what's *not* in question: dropping the dead grants.gov URL, counting a chrome-only page as broken so the existing sweep finds it, purging chunks before re-ingesting, and surfacing unmatched labels at import time are all right. It's the two detection mechanisms and the repair ordering that need another pass.

Happy to take any of these on if you'd rather not — say the word.

Issue #692, opened 2026-08-21
[Bug] Admin user-search boxes silently find nobody — GET /api/verification/examiners/search does not exist


arhyneRWU:
## Description

Two admin-only user-search boxes call `GET /api/verification/examiners/search`, and that route does not exist anywhere in the backend. Every request 404s. Both call sites `catch` the error and set their results to an empty array, so nothing surfaces the failure — the search box looks like it is working and reports that no user matches, whatever you type.

The two affected surfaces:

1. **Verification → Examiners → Search Users** — the only way to grant or revoke examiner status on a user who is not already an examiner. Since the list below the box shows only *existing* examiners, and the search is the only path to anyone else, **granting examiner status to the first/any new user is not possible through the UI at all.**
2. **Admin → Organizations →** (select an organization) **→ Add User** — the only in-app way to assign a user to an organization. Org membership drives access filtering (`organization_service`), so this blocks org-scoped access setup.

The backend service function is present and correct — `verification_service.search_users` at `backend/app/services/verification_service.py:1361`, which regex-matches `name`/`email` and returns exactly the `{user_id, name, email, is_examiner}` shape both callers expect. Only the router endpoint that would expose it was never added.

## Steps to Reproduce

**Examiner management (needs an admin account):**

1. Sign in as an admin.
2. Go to `/verification` and open the **Examiners** tab (admin-only; it does not render for non-admins).
3. Type any substring of a real user's name or email into **Search Users** and wait ~300ms for the debounce.
4. The dropdown shows **"No users found."** — for every query, including an exact email of a user who definitely exists.
5. With devtools open, the Network tab shows `GET /api/verification/examiners/search?q=… → 404`, and the console stays clean.

**Organization membership (needs an admin account):**

1. Sign in as an admin.
2. Go to `/admin` → **Organizations** tab.
3. Select any organization to open its members panel.
4. Type any substring of a real user's name or email into **Add User**.
5. The dropdown shows **"No results"** for every query. Same 404 in the Network tab.

## Expected Behavior

Typing part of a name or email returns matching users, each with an **Add** button (Organizations) or a grant/revoke examiner button (Examiners), so an admin can act on them.

## Actual Behavior

Every query 404s and both boxes render their empty state — "No users found." on the Examiners tab, "No results" on the Organizations tab. No toast, no console error, no indication that a request failed. The two are indistinguishable from a genuinely empty result set, which is why this has gone unnoticed.

## Screenshots

[...]

PR #693, opened 2026-08-21
feat(scripts): map every backend route to the frontend calls that hit it

arhyneRWU:
## Summary

Adds `scripts/map_ui_endpoints.py`, a standalone generator that answers two questions no existing tool in the repo can answer: **"is this backend route still called from the UI?"** and **"does this frontend call still have a backend route?"**

Static call graphs trace each half of the app precisely and then stop at the HTTP boundary. `apiFetch('/api/chat/truncate')` is a *string*, not a code reference, so nothing links it to `@router.post("/truncate")`. That gap is why an endpoint can be deleted while a call site keeps pointing at it, and why dead route surface accumulates unnoticed. The script builds exactly that missing edge and nothing else.

It works by importing the real FastAPI app and walking `app.routes` — so router prefixes, mount order, and feature-gated routers are *resolved*, not guessed by scraping `@router.get(...)` decorators — then parsing the `apiFetch` / `rawFetch` calls in `frontend/src/api/*.ts` for URL and method, and joining the two sides by URL shape (`{param}` and `${expr}` both normalize to a wildcard).

Run it from the repo root:

```
cd backend && uv run python ../scripts/map_ui_endpoints.py
```

It writes `scripts/ui_endpoint_map.md` and `scripts/ui_endpoint_map.json`. **Those outputs are generated on demand and not committed** — a checked-in copy goes stale the moment anyone adds a route, and a stale map is worse than no map.

### What it found on `main` (250160470bf0152888db6e634527708b43e4c2cc)

| | count |
|---|---|
| Backend routes | 525 |
| Frontend api-layer calls | 434 |
| Routes with no frontend caller | 96 |
| Frontend calls matching no route | 2 |
| API calls made outside `frontend/src/api/` | 4 |

**2 frontend calls match no backend route.** Both hit `GET /api/verification/examiners/search`, which does not exist anywhere in the app:

| Method | URL | Function | File:Line |
|---|---|---|---|
| GET | `/api/verification/examiners/search?q=…` | `searchUsersForExaminer` | `frontend/src/api/library.ts:379` |
| GET | `/api/verification/examiners/search?q=…` | `searchUsers` | `frontend/src/api/organizations.ts:146` |

These are live call sites behind two admin search boxes (Verification → Examiners, and Admin → Organizations → *Add User*). Both swallow the error and render "No results", so the failure is invisible — the boxes look like they work and simply never find anybody. The service function they want, `verification_service.search_users`, exists and is correct; only the route was never added. Filed separately as a bug report so it can be fixed independently of this PR.

**96 routes have no frontend caller.** Five of those are tagged `yes` in the *External-API doc?* column, meaning they appear in `docs/api.md` or `docs/mgmt-api.md` and are expected to have no UI caller. The report cross-tags them precisely so they read as intentional and the other 91 stand out for a human to triage — they are some mixture of legitimately headless surface (SAML/OAuth callbacks, Graph webhooks, browser-automation session control) and genuinely dead code. This PR does not delete anything; it just makes the list visible for the first time.

**4 API calls are made from outside `frontend/src/api/`**, which is a layering check the api layer cannot perform on itself:

- `GET /api/folders/all` — `components/workspace/AutomationCreationWizard.tsx:94`, `components/workspace/AutomationEditorPanel.tsx:510` and `:825`
- `GET /api/workflows/${…}/quality-sparkline` — `hooks/useQualitySparkline.ts:26`

### Scope, deliberately narrow

Anything *above* the api layer (which component calls which api function) and *below* the route (what a handler touches) is a call graph's job. This tool only spans the HTTP hop; compose the two for a full UI-to-database blast radius.

## Changes

- **`scripts/map_ui_endpoints.py`** (new, 414 lines, no new dependencies — stdlib plus the backend's own FastAPI app).
  - `scan_backend()` imports `app.main:app` and walks `app.routes`, capturing method, resolved full path, handler name, source file/line, and auth dependencies per route. `ENABLE_TRIAL_SYSTEM` and `TELEMETRY_COLLECTOR_ENABLED` are force-defaulted on for the walk so the map documents the full mountable surface; a deployment with them off simply does not serve those routes.
  - `scan_frontend()` parses `apiFetch` / `rawFetch` calls in `frontend/src/api/*.ts`. Template literals get real handling — a nested interpolation such as `` `/x/${q ? `?a=${a}` : ''}` `` parses instead of terminating at the first inner backtick — and generic type arguments (`apiFetch<Foo>(…)`) are skipped.
  - `scan_out_of_layer()` sweeps the rest of `frontend/src` for `fetch`/wrapper calls to `/api/…`.
  - The join normalizes both sides: a trailing interpolation glued to the last path segment (`…suite${qs}`) is treated as a query-string builder and dropped, while one following a slash (`/${id}`) is kept as a real path parameter.
  - `--check` exits 1 when any frontend call matches no backend route. `--stdout` prints the Markdown and writes nothing.
- **`CHANGELOG.md`** — entry under `[Unreleased] → Added`, matching how `scripts/cut_release.sh`, `scripts/dead-code-scan.mjs`, and the other operator/dev tooling additions were recorded.

### Not done here, on purpose

`--check` is **not** wired into CI. It currently exits 1, because of the two orphan calls above. A gate that fails the moment it lands gets bypassed rather than fixed, so the sensible order is: land the tool, fix the two call sites, then add the gate in a separate PR. Happy to do that follow-up if you want it.

### Known limitation

For the 17 routes decorated with `@limiter.limit(...)`, the *File* column reports slowapi's wrapper module rather than the real handler file, because `inspect.getfile()` follows the decorator's wrapper. Method, path, and handler name are correct for those rows; only the file attribution is wrong. Left as-is rather than special-casing one library's wrapper.

## Test Plan

Exactly what was and was not run, since two boxes are unticked:

**Ran:**

- `cd backend && uv run python ../scripts/map_ui_endpoints.py` — completes against current `main`; prints `routes=525 calls=434 uncalled=96 unmatched=2 out_of_layer=4` and writes both output files.
- `cd backend && uv run python ../scripts/map_ui_endpoints.py --check` — exits 1 with `CHECK FAILED: 2 frontend call(s) match no backend route`, confirming the CI-gate path works.
- `cd backend && uv run ruff check ../scripts/map_ui_endpoints.py` — clean. (Note: `make backend-lint` only covers `backend/app/`, so this file is outside CI's lint scope; ran it explicitly.)
- `cd backend && uv run bandit -s B101 -q -ll -ii ../scripts/map_ui_endpoints.py` — clean.
- Spot-verified the two orphan findings by hand against `main`: `grep`ed every `@router.*` decorator in `backend/app/routers/verification.py`, confirmed the router mounts at `prefix="/api/verification"` in `backend/app/main.py`, and confirmed only `GET`/`POST /examiners` exist — no `/examiners/search`. Also matched the path against every `APIRoute.path_regex` in the live app: zero matches.

**Did not run:**

- `make ci` — not run in full. This branch adds one standalone script under `scripts/` and one CHANGELOG line; it imports no application module into any test, and changes nothing under `backend/app/` or `frontend/src/`, so backend and frontend suites are untouched by it. Flagging rather than ticking a box I did not verify.
- `make release-check` — not run; packaging and deployment are unchanged by this PR.
- Deploy/release docs — not updated; the install and release paths are unchanged. Invocation is documented in the script's module docstring and in the header of every generated report.

## Related Issues

The two orphan calls reported above are filed as #692 (found by this tool; fixed independently — this PR does not close it).


arhyneRWU:
I built this mapping tool a few years ago when context limiters were quite
restrictive. It turns out to be really useful as the codebase gets larger — a
quick way to generate a map to understand the frontend and backend
interactions and to make sure we don't leave anything behind. I have updated
the script to work on this repo.

I suggest that we look at the implementation of this, along with the skill
that I use, which is graph code review, which does a really good job of
helping the agents understand the codebase and simplifying the view of the
codebase.

ViaJables (changes requested):
Reviewed, and I want this tool in-tree — it ran cold from a fresh worktree with no `.env` and every backing service down, reproduced your exact counts (`routes=525 calls=434 uncalled=96 unmatched=2 out_of_layer=4`), and it has already paid for itself by finding #692 (fix now open as #698). The import-time analysis holds up: `app.main` connects to nothing at import, so walking the live route table instead of scraping decorators is the right call.

Three small changes before merge:

1. **Dynamic-URL calls are silently dropped, and the comment says otherwise.** The `continue` for wrapper calls whose first argument isn't a literal is annotated *"reported via --check as unparsed"* — but nothing reports them: they're not counted, not listed, and don't affect `--check`. Two real call sites hit this today: `frontend/src/api/admin.ts:341` (`getWorkflowEvents`, builds `url` with `+=`) and `admin.ts:678` (`getQualityTimeline`). Consequence: `GET /api/admin/workflows` and `GET /api/admin/quality/timeline` land in "Routes with no frontend caller" even though the UI calls both — 2 of the 96 rows are false positives in exactly the list the tool pitches for dead-code triage. Please emit the skipped calls as their own count/section (there are only two), so a reader of the uncalled list knows its error bar.

2. **Gitignore the outputs.** The PR is emphatic that `ui_endpoint_map.md` / `.json` must never be committed, but nothing enforces that — they sit untracked waiting for a `git add -A`. Add `scripts/ui_endpoint_map.*` to `.gitignore`.

3. **Warn when the force-defaulted flags lose.** `setdefault` on `ENABLE_TRIAL_SYSTEM` / `TELEMETRY_COLLECTOR_ENABLED` is silently defeated by a shell that already exports them false, which contradicts "documents the full mountable surface". A one-line warning when the resolved value is falsy covers it.

Happy to merge as soon as those land.

On the graph-code-review skill: interested, but let's keep it out of this PR — open it as its own proposal (an issue with the skill's shape, like you did with #629/#628) and we'll look at it on its own merits.

arhyneRWU:
All three landed in 166a5e39 (plus a merge of current `main`):

1. **Dynamic-URL calls are now surfaced, not dropped.** They get their own summary count, a "Calls with runtime-built URLs" section in the Markdown (placed right above the uncalled-routes list and labeled as its error bar), and a `dynamic_url_calls` list in the JSON. The wrong "reported via --check" comment is gone. One wrinkle worth noting: naively yielding every non-literal first argument also caught the `apiFetch`/`rawFetch` *definitions* in `client.ts` (the parameter list parses like a dynamic call), so declarations are excluded — the list is exactly the two real call sites you named, `getWorkflowEvents` (admin.ts:341) and `getQualityTimeline` (admin.ts:678).

2. **Outputs gitignored** — `scripts/ui_endpoint_map.*` added to `.gitignore`.

3. **Defeated-flag warning** — after the app import, the script checks the *resolved* `Settings` values (so a falsy shell export and a `.env` are both caught) and warns on stderr per flag that its routes are missing from the map.

Fresh run on the merged tree: `routes=526 calls=434 uncalled=96 unmatched=0 dynamic=2 out_of_layer=4` — `unmatched` dropped 2→0 because #698 added the route both orphan calls wanted, and `routes` went 525→526 for the same reason.

On the graph-code-review skill: agreed, keeping it out of this PR — will open it as its own proposal.

ViaJables (approved):
Re-verified all three changes on 9440f6a9 — dynamic-URL calls surfaced with their own section/JSON key (ran it: exactly the two admin.ts sites), outputs gitignored, and the setdefault warning fires when the shell exports the flags false. One nit for later: the warning text says `.env` can defeat the default, but only a process-env export can (pydantic-settings prefers process env). Merging.

Issue #704, opened 2026-08-21
[Feature] Review tooling behind the Makefile: endpoint-map-check in CI, plus an opt-in local code graph


arhyneRWU:
## Problem

Reviewing a change in this codebase means reconstructing structure by hand:
which callers a changed service function has, whether anything tests it, what
a rename would touch. At 900+ files across Python and TypeScript that gets
re-done from scratch every review, and for agent-assisted review it is also
the dominant context cost — answering "who calls this?" by opening files pulls
in far more than the answer needs.

There is also a narrower, sharper gap. `scripts/map_ui_endpoints.py` (#693)
exits non-zero when a frontend call matches no backend route — the #692 bug
class — and nothing runs it. In #693's CHANGELOG I argued against wiring it
into CI because "a gate that fails on arrival gets bypassed." That objection
has expired: with #698 merged, `unmatched` is **0**, so the gate would go in
green.

This is the follow-up you asked for in the #693 review. It comes in two
independent halves — please feel free to take the first and decline the
second.

### Half 1: make the endpoint-map check enforceable (small, self-contained)

Two Makefile targets, and one CI step:

```make
endpoint-map: ## in the hand-written help block, matching this file's style
	cd $(BACKEND_DIR) && uv run python ../scripts/map_ui_endpoints.py

endpoint-map-check:
	cd $(BACKEND_DIR) && uv run python ../scripts/map_ui_endpoints.py --check --stdout >/dev/null
```

`--stdout >/dev/null` matters: `--check` alone still writes
`scripts/ui_endpoint_map.md` and `.json`, both timestamped, so a bare
`--check` dirties the tree on every run. `--stdout` suppresses the write and
still returns the right exit status.

Then hang it off the target that already exists, rather than adding a bespoke
CI step:

```make
backend-static: backend-lint backend-security endpoint-map-check
```

That's the whole CI change — **no workflow YAML at all**. The "Backend Release
Checks" job already runs `make backend-install` then `make backend-static`, so
the check rides a job that's already installing exactly what it needs, at a
marginal cost of ~2 s.

**Proposed as a gate, not advisory.** The repo has the `continue-on-error`
ramp for `backend-typecheck` and `backend-audit`, but that pattern exists for
checks with a pre-existing backlog to work down. This one is at zero today, so
there is nothing to ramp: an advisory gate on a green check only teaches
people to scroll past it. The cost of the gate is two seconds and the failure
mode it prevents already shipped once.

[...]

PR #706, opened 2026-08-21
build: optional review-graph target for a local code graph (part of #704)

arhyneRWU:
> Half 2 of #704, filed so you can read the diff rather than the description
> before deciding. **Close it freely** — it is a standalone convenience, and
> #705 (Half 1, the endpoint-map gate) does not depend on it. Branches off
> `main`, no stacking.

## Summary

Adds `make review-graph`, an optional local code graph for review navigation,
plus its docs. It answers structural questions — callers, callees, tests-for,
impact radius, affected flows — without opening files, using
[code-review-graph](https://github.com/tirth8205/code-review-graph) (MIT, on
PyPI, third-party; I use it, I'm not the author).

**Opt-in by construction rather than by flag.** No other target depends on it,
nothing in CI touches it, it is in neither lockfile, and the recipe is the only
thing in the repo that mentions the tool. A flag can be switched on by
accident; a target nothing calls cannot be. If you take nothing else from this
PR, the `.gitignore` commit stands on its own.

If the tool is not installed, the target prints the install command and exits
non-zero rather than skipping quietly:

```
$ make review-graph
review-graph: code-review-graph is not installed.
  install it with: uv tool install code-review-graph
  it is optional -- no other make target needs it.
```

A silent skip would be the wrong failure mode for a command someone typed —
and it is the specific trap this tool sets, which brings me to the two
corrections in the recipe.

## Two bugs corrected in the recipe rather than inherited

Both found while preparing #704, and both are why this is 25 lines of Makefile
rather than two:

1. **`update` does not build.** On a repo with no graph, `code-review-graph
   update` creates the database, indexes `HEAD~1..HEAD`, and **exits 0** —
   leaving a graph of about one file that answers every subsequent query
   confidently and wrongly. Reproduced on a throwaway 3-commit repo:
   `Nodes: 2, Files: 1`, exit 0. The target therefore tests for
   `.code-review-graph/graph.db` and calls `build` explicitly rather than
   relying on `update || build`, which never falls through.
2. **`--base` defaults to `HEAD~1`.** A refresh after pulling several commits
   would index only the last one, silently. The target diffs from `ORIG_HEAD`
   (which git sets across a pull), falling back to `HEAD~1` when it is unset,
   and honours `CRG_BASE` for anything else.

## Changes

- **`Makefile`** — `review-graph` target, added to `.PHONY` and to the
  hand-written `help` block. Not a prerequisite of anything.
- **`docs/review-graph.md`** — setup, refresh semantics, and a "what it gets
  wrong" section that is deliberately longer than the setup section:
  - dynamic dispatch produces no edges (Celery `.delay()`/`apply_async`,
    `Depends()` injection, Beanie's document-list registration, string-registered
    routers, runtime-built pydantic-ai agents), so `callers_of` can be silently
    incomplete;
  - there are no route-to-handler edges — that is `scripts/map_ui_endpoints.py`'s
    job (#693), and the two compose rather than overlap;
  - `TESTED_BY` is heuristic;
  - semantic search needs a multi-GB extra and is not in the base install;
  - the tool ships optional embedding backends (MiniMax, OpenAI-compatible,
    Google) that would send code-derived text to a third-party API — inert
    without an API key, and the docs say to leave them off, since the graph is
    built from a working tree that on this product may sit beside unpublished
    proposals.
  - It also documents `code-review-graph install --no-instructions`, because
    without that flag the tool's installer edits `CLAUDE.md`, which is tracked.
- **`CONTRIBUTING.md`** — four lines noting the target exists and is not part
  of any check command.
- **`.gitignore`** — `.mcp.json`, `.cursor/`, `.kiro/`, `.qoder/`. These are
  per-developer AI-tool config that installers write into the repo root; today
  they sit untracked, one `git add -A` from being committed. **Own commit
  (d47e943) and independently useful** — worth taking even if you close the
  rest. One judgement call to flag: some projects intentionally commit
  `.mcp.json` to share server config with the team. If you might want that,
  drop that line and keep the three editor directories.
- **`CHANGELOG.md`** — entry.

## Test Plan

- [ ] Release check passes if packaging or deployment changed
      (`make release-check`) — not run: nothing here touches packaging or
      deployment, and no target added here is reachable from it.
- [ ] Updated deploy/release docs — n/a, install and release paths unchanged.

Every branch of the recipe was exercised, using a stub on `PATH` so the
dispatch is verified rather than assumed:

| Case | Result |
|---|---|
| Tool not on `PATH` | prints install instructions, exit 2 |
| No `graph.db` present | calls `build` (not `update`) |
| `graph.db` present, `ORIG_HEAD` unset (fresh clone) | `update --base HEAD~1` |
| `graph.db` present, `ORIG_HEAD` set (after a pull) | `update --base <pre-pull sha>` |
| `CRG_BASE=main make review-graph` | `update --base main` |
| `make help` | lists the target |

## Related Issues

Implements Half 2 of #704. Independent of #705 and of #693.


arhyneRWU:
Rebased onto current `main` and taken out of draft, so this is a real yes/no rather than something still in progress. I suspect the draft status is why it has been quiet — that is on me.

**Nothing here has changed since you closed #704.** Same three commits, same +123/-1 across five files. The rebase had three conflicts, all purely additive, and one of them is worth naming because resolving it the obvious way would have been a regression:

- **`Makefile`** — this branch predates #705, so its version of the target line reads `backend-static: backend-lint backend-security`. Taking that side would have silently deleted the `endpoint-map-check` gate you merged. Kept yours; the only change to that line now is `review-graph` added to `.PHONY`. Verified: `backend-static` is byte-identical to `main`, and `make endpoint-map-check` still passes.
- **`.gitignore`** and **`CHANGELOG.md`** — kept both sides, entry added at the top of Added.

**On the decision itself — closing this is a perfectly good answer**, and I would rather you close it than merge it out of politeness. You took Half 1 and closed #704 as completed, which reads as a no to Half 2, and I would not disagree with that call: it is a standalone convenience with a third-party dependency, and nothing else needs it.

Two things worth weighing if you are on the fence:

1. **The `.gitignore` commit (`c2340a29`) is useful whether or not you take the graph** — it ignores per-developer AI/editor config that tools write into the repo root. Happy to split it into its own two-minute PR and close this one, if that is the shape you want. One judgement call still stands from the original body: some projects deliberately commit `.mcp.json` to share MCP config with the team, so drop that line and keep the three editor directories if that is you.
2. **It cannot be switched on by accident.** No target depends on it, nothing in CI touches it, it is in neither lockfile, and the recipe is the only mention of the tool in the repo. If the tool is not installed the target prints the install command and exits non-zero rather than quietly doing nothing.

No rush and no hard feelings either way — a one-word answer closes it out.

PR #713, opened 2026-08-24
fix(extraction): record whether a source quote supports its value

ViaJables:
Phase 1 of tightening the per-field source badge. Ships the correctness work and the unambiguous bug fix; deliberately does **not** change what the badge shows yet.

## Problem

`verified` on a field's source sidecar attests only that the quoted passage exists in the document (`extraction_sources.py`, `verified = offset is not None`). It says nothing about whether that passage supports the value shown beside it.

So: a model invents an award amount, returns any genuine sentence from the budget section as its supporting quote, the quote is found, and the field renders with a page badge and click-to-highlight — the strongest trust signal the product has, earned by a hallucination. For a grants office acting on an extracted dollar amount or deadline, that is the failure that matters most.

## What this changes

**Adds the missing check as a separate signal.** `resolve_entity_sources` now records `value_supported` and `value_support_method` per field — does the extracted value actually appear in its quote — via three deterministic paths:

| path | catches |
|---|---|
| literal | verbatim copy, after the existing unicode folding |
| numeric | `$61,100` in the value vs `61,100` in the prose |
| date | `March 1, 2026` vs `3/1/2026` |

No LLM, no new dependencies. The number/date parsers are written locally rather than imported from `extraction_validation_service` so the module keeps its documented pure-string contract (importing it would pull in the engine, Beanie documents, and system config, and invert the dependency direction).

**`verified` is unchanged, and no UI keys off the new field.** That is the point of shipping it this way — see below.

**Fixes `_backfill_sources` outright.** It copied pass 1's quote onto pass 2's value without checking the value survived refinement. Pass 2 routinely *changes* values, so a field could display "April 1" beside a real, verifiable passage reading "…due March 1…" — a source badge pointing at evidence against itself. Backfill now requires the values to match, formatting-tolerantly, so a reformatted-but-identical value still gets its source.

## Why the badge doesn't move in this PR

Nobody knows what fraction of real fields would flip from green to amber. That number decides the design, and guessing it wrong is worse than the current too-generous badge: if amber appears on half the fields of a correct extraction, users learn to ignore amber within a week and the signal is spent.

The check therefore returns **not applicable** rather than `false` for the classes that would load the measurement with non-hallucinations — enum fields (the value maps the prose rather than quoting it), "not found" sentinels, non-scalar values, and quotes that were never located in the first place (a fabricated passage containing the value proves nothing).

**To measure once this has been running:** the sidecar is persisted in `activity_event.result_snapshot.sources` for run-sync extractions. Group by `value_support_method` to get the distribution — `no_match` is the population that would turn amber, and `enum_field` / `empty_value` / `non_scalar_value` are the ones correctly excluded. A high `no_match` rate concentrated in a few templates most likely means those templates have judgment fields missing `enum_values`, which is a metadata fix, not a badge design problem.

Two known weaknesses are documented in the docstring so the distribution is read correctly: very short values (`"1"`) match almost any passage coincidentally, and a judgment value (`"Yes"` for "cost sharing is required") reads as unsupported unless the field declares `enum_values`. An existing test in this file is exactly that second case, and I left it asserting `false` with a comment rather than papering over it — it is the clearest illustration of why measurement comes before the badge.

## Follow-ups (not in this PR)

- **Phase 2:** measure the distribution.
- **Phase 3:** redefine what earns the strong badge, and give the sidecar's four real states two visual tiers (rely on it / check it yourself). That pass should also fix the related defect the audit found: a field where the model returned *no* quote at all currently renders completely unmarked, so the most uncertain state looks the cleanest.

## Tests

New `TestValueSupport`, `TestBackfillGuard`, `TestSameValue` in `tests/test_extraction_sources.py`, including the headline case — hallucinated `$500,000` paired with a genuine budget sentence now yields `verified: True, value_supported: False` — plus currency/date equivalence, the not-applicable classes, and both backfill directions.

Locally: `test_extraction_sources.py` (38) and `test_extraction_engine_core.py` (46) pass, `ruff check app/` clean. Only that one test file references the sidecar anywhere in the suite. The full suite is left to CI here — this machine is currently saturated by another job and local full-suite runs were taking >10min.

### DIFF
--- a/backend/app/services/extraction_engine.py
+++ b/backend/app/services/extraction_engine.py
@@ -544,6 +544,12 @@ class ExtractionEngine:
                 if field not in entity or field in sidecar:
                     continue
                 if not _same_value(entity.get(field), draft_value):
+                    # Withhold the quote, but leave an entry behind. Dropping
+                    # the field entirely makes the UI render it unmarked —
+                    # the state that looks cleanest — so a revised value would
+                    # lose both its badge and its warning, and would vanish
+                    # from the value-support distribution instead of counting.
+                    sidecar[field] = {"quote": None, "dropped_reason": "value_changed"}
                     continue
                 sidecar[field] = src
 
--- a/backend/app/services/extraction_engine.py
+++ b/backend/app/services/extraction_engine.py
@@ -17,7 +17,11 @@ from pydantic import BaseModel, ConfigDict, Field, create_model, model_validator
 from pydantic_ai import Agent, BinaryContent, NativeOutput
 
 from app.models.system_config import DEFAULT_EXTRACTION_CONFIG, _deep_merge
-from app.services.extraction_sources import SOURCE_KEY, resolve_entity_sources
+from app.services.extraction_sources import (
+    SOURCE_KEY,
+    resolve_entity_sources,
+    same_value as _same_value,
+)
 from app.services.llm_service import (
     build_thinking_model_settings,
     create_chat_agent,
@@ -202,7 +206,7 @@ class ExtractionEngine:
                     self._resolve_sources(
                         doc_results,
                         texts[idx] if idx < len(texts) else "",
-                        (doc_metadata or []), idx,
+                        (doc_metadata or []), idx, meta_map,
                     )
                 all_results.extend(doc_results)
             return all_results
@@ -222,15 +226,18 @@ class ExtractionEngine:
                 capture_sources=capture_sources,
             )
             if doc_results and capture_sources:
-                self._resolve_sources(doc_results, doc_text, (doc_metadata or []), idx)
+                self._resolve_sources(doc_results, doc_text, (doc_metadata or []), idx, meta_map)
             all_results.extend(doc_results)
 
         return all_results
 
     @staticmethod
-    def _resolve_sources(entities: list, doc_text: str, doc_metadata: list[dict], idx: int) -> None:
+    def _resolve_sources(
+        entities: list, doc_text: str, doc_metadata: list[dict], idx: int,
+        field_meta: dict[str, dict] | None = None,
+    ) -> None:
         meta = doc_metadata[idx] if idx < len(doc_metadata) else {}
-        resolve_entity_sources(entities, doc_text or "", meta or {})
+        resolve_entity_sources(entities, doc_text or "", meta or {}, field_meta)
 
     def build_from_documents(self, doc_texts: list[str], model: str) -> dict | None:
         """Generate extraction entities from document text using LLM."""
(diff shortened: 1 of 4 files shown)


ViaJables:
Reviewed and pushed fixes as `53d88140`. The design here is right — separating "the quote exists" from "the quote supports the value", and refusing to answer for the four classes that would pollute the measurement. Leaving the "Yes"/cost-sharing test asserting `False` with an explanation, rather than quietly adjusting it, is what made the rest of the diff easy to trust.

Four things needed fixing first, all reproduced by running the code.

**1. The literal path had no boundaries**, so it certified values the quote contradicts:

| value | quote | was |
|---|---|---|
| `500000` | "The total award is **5000000** dollars…" | `(True, literal)` |
| `1,500,000` | "The ceiling is $1**1,500,000**…" | `(True, literal)` |
| `$500,000` | "The obligated total is **$500,000**.99…" | `(True, literal)` |
| `61100` | "Account code 5**61100** covers equipment…" | `(True, literal)` |
| `John Smith` | "The contact is **John Smith**son, Director…" | `(True, literal)` |

Matches must now sit on value boundaries, and every occurrence is scanned — the first may be a fragment (`500,000` inside `1,500,000`) while a later one is genuine. A trailing `.` or `,` only counts as part of a number when a digit follows, so `"The award is $500,000."` still resolves as supported.

**2. Number parsing conflated units and swallowed lists.** `%` sat in `_CURRENCY_CHARS`, so a `$50` value was supported by *"50% of MTDC"*; the percent now travels with the number. `[\d,]*` read `"Sections 1,2,3"` as the single number `123`.

**3. No quote length cap**, against a date scan that backtracks quadratically on space-free text — 6KB → 715ms versus 7.6ms for normal prose of the same size. The quote is model-controlled and `capture_sources` runs in-request, so quotes past 4000 chars are excluded as `quote_too_long`. Worth noting this governs four call sites once it merges forward to `major/agentic-chat`, not one.

**4. Withholding a quote dropped the sidecar entry entirely.** At `ExtractionEditorPanel.tsx:1601-1604`, an undefined `src` makes both `hasSource` and `noSource` false, so a value pass 2 revised lost its badge *and* its warning — the least trustworthy state rendering as the cleanest — and vanished from the distribution instead of counting. The entry is now kept with the quote withheld, resolving as `value_changed_in_refinement`.

Note the fix belongs in the backend rather than the badge component: the sidecar has one shape but three provenance guarantees (live run, replayed chat, editor history), and a missing entry means "this field got no quote" in the first and "this run captured nothing" in the others. A frontend-only `noSource = !hasSource` would light the amber warning on every field of any pre-source-tracking run. The equivalent bug shipped in chat and was fixed separately in `aecc65ce`.

**One limitation the boundaries can't fix**, now documented in the docstring: the check asks whether the value appears *somewhere* in the passage, not whether the passage assigns it to this field. A "Direct Costs" value is still supported by a quote labelling that figure as indirect, and a year (`2024`) by a dollar amount (`$2,024`). Reading a `True` as "the quote says this field is this value" over-reads it — worth stating before the Phase 2 numbers get interpreted.

**Also worth correcting in the description:** "does not change what the badge shows" isn't accurate — `_backfill_sources` changes it for revised values, as above.

Tests: added `TestLiteralMatchBoundaries` (both directions), `TestNumberParsing`, `TestQuoteLengthCap`, and an end-to-end case for the withheld entry surviving resolution as present-but-unverified. Updated `test_changed_value_does_not_inherit_the_drafts_quote`, which pinned the old drop-entirely behavior. 106 tests in the two files, 627 across all extraction/source tests, `ruff check app/` clean.

PR #726, opened 2026-08-26
fix(kb): per-source Refresh for URL sources, honest add-URL counts, seed_catalog --refresh-urls

ViaJables:
## Summary
Support ticket: the catalog OSP KB (`7899fe74…` on vandalizer.uidaho.edu) still served the **December 2018** text of APM 45.13 / 45.14 after the University revised both pages in July 2026, and the author's refresh attempt — re-adding the two URLs, then re-exporting — changed nothing. It couldn't: `add_urls` skips URLs the KB already holds, the route reported `added: 2` from the request length, and the UI toasted "Added 2 URLs — crawling in background" on a no-op. Nothing else fetched either: **Rebuild index** re-embeds the stored snapshot, and a catalog upgrade only adds unseen sources. (The deployed KB's `updated_at` is 2026-06-24, so the re-add never even reached it.)

- **Per-source Refresh** — `knowledge_service.refresh_url_source` re-fetches a URL source in place (same uuid/label/provenance) through the same content gates as first ingest (extracted into shared `_reject_fetched_page`). A fetch that fails or returns a bot-challenge/boilerplate page **keeps the previous text and chunks** and records why on the source. Exposed as `POST /knowledge/{uuid}/source/{source_uuid}/refresh` → Celery `tasks.kb.refresh_url_source`; ↻ button on URL rows in KnowledgePanel with the last-fetched date (`processed_at` now on the list response).
- **Honest add-URL counts** — `partition_new_urls` runs before dispatch; the response returns `added` / `skipped` / `skipped_urls`, only new URLs go to the worker, the KB isn't left `building` when nothing was queued, and the toast says "N URLs already in this KB — nothing was fetched" and points at Refresh.
- **`seed_catalog --refresh-urls`** — re-fetches existing live-URL sources of already seeded KBs (bundled `content_file` sources are never refetched; failures keep the old text). Also fixes the stale APM 45.14 `url_title` in the seed.
- CHANGELOG entry under Unreleased → Fixed.

## Test plan
- [x] `ruff` clean on touched files
- [x] 7 new service tests (`tests/test_kb_source_refresh.py`), 4 new route tests; `test_knowledge_routes`, `test_kb_source_dedup`, `test_kb_crawl`, `test_knowledge_base_tasks`, seed/bootstrap tests all pass
- [x] `tsc` clean, eslint 0 errors, `KnowledgePanel.test.tsx` 9/9 (3 tests updated/added)
- [ ] After deploy: ↻ on APM 45.13 and 45.14 in the OSP KB (or `seed_catalog --only kbs --refresh-urls`), confirm the export shows `Last updated: July 2026` / `July 13, 2026`, then close the ticket


ViaJables:
Review notes (merging as part of today's batch):

- **Fixed on the branch:** `backend/seeds/VERSION` bumped 1.3.2 → 1.3.3 — the seed-version-check job was failing because `ui_sponsored_programs.json` changed (APM 45.14 `url_title`) without a bump, so existing installs would never have seen the catalog update.
- Verified: 186 backend tests green across the KB/catalog/bootstrap files, `tsc` clean, `KnowledgePanel.test.tsx` 9/9, endpoint map has no orphans (`refreshKBSource` ↔ the new route).

Follow-ups worth a ticket, none blocking:
1. `knowledge_service.refresh_url_source` deletes the old chunks (`delete_kb_source`) *before* `add_to_kb`; if re-embedding fails, the source keeps its old `content` but its chunks are gone and status goes `error`. The "previous content kept" guarantee only covers fetch/gate failures. Swapping the order, or re-adding from `source.content` on failure, would make it hold for re-indexing too.
2. `kb_validation_tasks.refresh_url_source_task`: if `svc.refresh_url_source` raises after status was set to `processing`, `recalculate_stats` is skipped → KB stuck `building`, source stuck `processing`, and the router's 409 guard then blocks further Refresh on it. Same shape as `_add_urls_async`; a `try/finally` around the recalc is cheap.
3. The 409 guard checks `processing` but not `pending`, so two rapid API POSTs queue two tasks (UI disables on `pending`, so API-only).
4. Refreshing a crawl parent doesn't re-crawl children, and a crawled child in `skipped` status is a URL source whose ↻ bypasses the crawler's `content_gate` — it would embed a nav page. `types/knowledge.ts` status union lacks `'skipped'`, so the button is enabled for those rows. Hiding ↻ on `parent_source_uuid` rows or applying the gate would close it.

PR #739, opened 2026-08-27
fix(llm): back off on 429 instead of the SDK's three-in-400ms retry

ViaJables:
## Summary
- Sentry **VANDALIZER-BACKEND-2T**: `ModelHTTPError 429 — Rate limit exceeded: 200 requests per minute` from `tasks.kb.optimize_kb` on prod. The gateway cap is per key per minute and shared by every caller on the deployment; the optimizer's answer + judge fan-out (4 in flight, two models on one key) plus everything else pushed it to the ceiling.
- The OpenAI SDK's built-in retry is 2 retries ~0.4 s apart — the breadcrumbs show all three attempts inside 400 ms. Against a per-*minute* window that's three knocks on the same closed door.
- New `RateLimitRetryTransport` under the shared per-loop httpx client (`_get_loop_http_client`, which every provider path uses — OpenAI-compatible, OpenRouter, Ollama/vLLM, Gemini): honours `Retry-After` (delta or HTTP-date, capped 60 s), otherwise exponential backoff with jitter (2/4/8/16/30 s ≈ a minute total), logs each wait at warning. Only 429 is handled; 5xx/connection errors keep the SDK's short retry.
- On exhaustion it **returns** the real 429 rather than raising — pydantic-ai's own `AsyncTenacityTransport` re-raises, which the OpenAI SDK would re-wrap as a misleading "Connection error". Callers keep seeing `ModelHTTPError(429)` and their existing handling.
- 5 transport tests (retry-then-success with exact waits, Retry-After honoured, exhaustion returns 429, non-429 passthrough, loop client built on the transport); CHANGELOG under Unreleased.

Not changed: the optimizer's concurrency (4). The backoff self-regulates; if tuning runs keep starving interactive chat, lowering that is the next lever.

## Test plan
- [x] `uv run pytest tests/test_llm_service.py tests/test_kb_validation_tasks.py tests/test_kb_validation_service.py tests/test_metering.py` — 127 passed
- [x] `ruff check` clean on changed files
- [ ] CI


ViaJables:
Review notes (merging as part of today's batch):

- **CHANGELOG wording corrected on the branch.** The SDK's default is 2 retries with 0.5 s initial delay and jitter, so its three attempts span ~1–1.5 s, not 400 ms; and the new transport sits *under* those SDK retries rather than replacing them, so each SDK attempt runs the full 2/4/8/16/30 s loop — worst case ≈18 gateway requests and ~3 min per model call with no `Retry-After` (≈17 min if the gateway sends `Retry-After: 60`). Celery doesn't multiply it (`ModelHTTPError`/`RateLimitError` are excluded from `TRANSIENT_EXCEPTIONS`). The bullet now says "per SDK attempt" and mentions the stacking.
- **Follow-up, not blocking:** passing `transport=` to `httpx.AsyncClient` makes httpx skip `HTTPS_PROXY`/env proxies (`allow_env_proxies = trust_env and transport is None`). Nothing in `.env.example`/`DEPLOY.md` documents a proxy, so likely nobody is affected, but a deployment that needs one would lose it silently. `httpx.AsyncHTTPTransport(proxy=...)` from env would close it.
- Consider `max_retries=0/1` on the SDK clients for 429 specifically so the ceiling is the transport's alone — the optimizer's 4-way fan-out is 72 in-flight retries against a 200 rpm cap in the worst case.

PR #741, opened 2026-08-27
fix(workflows): initialise Beanie on the loop that sends approval-request emails

ViaJables:
## Summary
- Sentry **VANDALIZER-BACKEND-2V** (`CollectionWasNotInitialized` — "Failed to persist EmailLog", preceded by "Failed to apply email branding") from `tasks.workflow.resume_after_approval`.
- `_notify_approval_reviewers_sync` sends the reviewer email from a sync Celery task on a throwaway event loop. `send_email` touches Beanie twice on that loop — `SystemConfig.get_config()` for branding and `EmailLog.insert()` for the audit row — and Motor clients are loop-bound, so with no `init_db` on that loop both fail every time. The mail still goes out (SMTP is fine), but unbranded, unlogged, and with two Sentry errors per reviewer — on the initial run and on resume alike.
- Fix: the send now runs through `run_task_async` with `init_db(settings, skip_indexes=True)` first, matching how every other Celery task that touches these models does it (retention, engagement, approval sweeper).
- Regression test asserts `init_db(skip_indexes=True)` runs before `send_email` on that loop; CHANGELOG under Unreleased.

## Test plan
- [x] `uv run pytest tests/test_workflow_approval_pause.py tests/test_workflow_tasks.py` — 66 passed
- [ ] CI


ViaJables:
Review notes (merging as part of today's batch) — two commits added on the branch:

- **`init_db` once per notification, not once per reviewer.** The per-reviewer `_send()` closure built a Motor client (minPoolSize=10) and an event loop per assignee; the loop now renders every email first, then one `run_task_async` does a single `init_db` and sends them all, with a per-recipient try/except so one SMTP failure doesn't skip the rest. Matches `approval_tasks.py`. New test `test_init_db_runs_once_for_many_reviewers` covers the count and the isolation.
- **Test isolation.** `test_workflow_tasks.py`'s four approval-pause tests didn't mock `init_db`/`send_email`, so after this change each paid the 5 s Mongo server-selection timeout per Beanie call — the file went from 6.7 s to ~49 s, and stalled for minutes when run in one process with the other task tests. Added a module-level autouse fixture mocking both; back to ~3 s.
- Noted, unchanged: if `init_db` itself raised, the mail would now not go out at all (before, it went out unbranded). Low risk with `skip_indexes=True` and lazy Motor construction.

PR #764, opened 2026-08-28
fix(extraction): a key-mismatched or all-null run must not report "not found"

ViaJables:
Part 1 of 10 from the production-credibility review. **Extraction trustworthiness.**

A grants officer acting on a wrong extracted deadline or dollar amount is a career-level failure, and the single most dangerous thing this product can do is render a failure as `not in the document`. Two paths still did that.

## The default first pass could invent a page of nulls

`DEFAULT_EXTRACTION_CONFIG` sets `two_pass.pass_1.structured: False`, so `_extract_fallback_json` is the default first pass — not an edge case. It projected the parsed payload onto the requested keys with an exact `parsed.get(key)`:

```python
entity = {key: parsed.get(key) for key in keys}
```

A model answering `"Award Amount"` for the requested key `"Award amount"` therefore produced `{"Award amount": None, ...}` for every field. Downstream that is indistinguishable from "none of these fields appear in the document".

Now: keys reconcile on a folded form (case, whitespace, punctuation), exact matches still win over folded ones, the `_sources` block is reconciled the same way so quotes are not silently lost, and a payload that shares **zero** keys with the request raises `ExtractionError` rather than returning nulls. A model that answers with the right keys and null values is untouched — absence stays absence.

## An all-null run was recorded as completed

`routers/extractions.py` guarded with:

```python
no_values = not any(isinstance(e, dict) and e for e in results)
```

`bool({"a": None})` is `True`, so only a *zero-entity* result was caught. An all-null entity sailed through as `COMPLETED` — a green tick in History over answers that were never successfully looked for. The check is now "at least one real value", skipping the `_field_sources` sidecar, which is metadata about the values rather than a value.

## Tests

`test_extraction_fallback_key_drift.py` (8 cases) and `test_extractions_all_null_run.py` (5 cases). Existing engine, prompt-variant and no-fields suites pass unchanged — 75 passed.

## Branch choice

Both files are byte-identical between `main` and `major/agentic-chat`, so this lands here and flows into v5 on the next catch-up merge.


ViaJables:
## Review round 1 — the guard was firing on the correct answer

`/code-review 764 high` found that this PR's central change raises on the payload its own prompt asks for. Verified end to end:

1. All three variants in `PROMPT_VARIANTS` end with *"Return a JSON object with an `entities` key containing a list of extracted objects."*
2. `_extract_fallback_json` uses exactly those prompts (`_resolve_prompt`).
3. So the **obedient** answer is `{"entities": [{…}]}`, whose only top-level key folds to none of the requested fields.
4. The new zero-match check raised on it.
5. The structured path has handled this all along via `coerce_entities`; the fallback path had no equivalent.

That envelope is very likely the original bug's actual cause — the old `parsed.get(key)` projection turned it into an entity of all-nulls, which is precisely the "confident not-found" this PR set out to stop. Raising on it trades a silent wrong answer for a hard failure. Neither one extracts.

**Fixed in e239c429**, along with two adjacent holes the same review surfaced:

- The list branch's "keep extra keys" loop re-attached `_sources` as a *data field*. `_entity_has_values` skips only `_field_sources`, so a non-empty `_sources` dict counted as a real value and an all-null run finished COMPLETED — this PR's own failure mode, straight back through the list path. Also meant list payloads never got their quotes reconciled into `SOURCE_KEY`.
- `/run-integrated` still finished COMPLETED unconditionally, so the run `/run-sync` now fails was still a green tick for API integrators.

8 new tests pin the envelope cases (including quotes-on-the-envelope, the quote block never becoming a field, and a genuinely off-topic answer still failing). 4134 passed.

### Not addressed — flagging for a decision

The review also noted the **blast radius of the raise**: `extract()` loops documents with no per-document `try`, so one document out of twenty whose pass-1 reply is genuinely off-topic now aborts the whole run and loses the other nineteen documents' values. With the envelope fixed that case is rare, and failing loudly is this PR's whole thesis — but "partial success with the bad document marked" is a different design call and I did not want to make it silently. Worth deciding before this lands.

PR #765, opened 2026-08-28
feat(extraction): run cross-field rules on every run, not just in the Validate tab

ViaJables:
Part 2 of 10 from the production-credibility review. **Extraction trustworthiness.**

> **Stacked on #764** (same function in `routers/extractions.py`). Base is `fix/extraction-null-entity-runs`; retarget to `main` after #764 merges.

Cross-field rules are the only checks in this product that can say a number is **wrong**, as opposed to merely untraced. A budget whose parts do not sum to its total. An end date before a start date. A required field missing given the condition that requires it. Exactly the failures a sponsored-programs office cannot afford to pass through.

They were reachable from two places: the explicit `POST /search-sets/{uuid}/validate-cross-field` endpoint, and the optimizer. `run-sync` never called them. So every production extraction shipped with none of them applied, and the check only happened if the user knew a separate button existed and remembered to press it.

## Change

`run-sync` evaluates the set's rules over the run's merged values, returns `{results, summary}` alongside the values, and persists them into the activity snapshot so reopening a run restores them. The panel renders the outcome above the extracted fields, failures in red, with each rule's message.

## Three deliberate constraints

- **Read-only.** `run_validation` calls `apply_evaluation_counters`, which feeds the false-positive auto-disable ratio. Production runs deliberately do **not**, so a rule cannot be pushed toward auto-disable by traffic the user never reviewed, and it avoids a `SearchSet` write on every extraction.
- **Absent is not passing.** `cross_field` is `null` when the set defines no rules. Absent means "nothing was checked", never "everything passed".
- **`unparseable` stays visible.** A rule the parser could not evaluate is shown as such rather than folded into the pass count. The tri-state in `cross_field_validation.py` exists precisely so a parser blind spot is not scored as either outcome.

Evaluation never raises: a rule-engine fault must not fail an extraction that already produced values.

## Tests

`test_cross_field_on_every_run.py` covers no-rules, a failing budget, a passing budget, unparseable-is-neutral, and engine-fault containment. Frontend `tsc --noEmit` clean.


ViaJables:
## Review round 1 — three ways the strip claimed more than the data supports

`/code-review 765 high` found six issues; four are fixed in bfa4ba56.

**1. Green "All 0 cross-field checks passed".** `failed` was false whenever no rule returned `fail`, so a report of 3 unparseable and 0 decisive rules rendered in the **success palette**. A budget whose fields come back as "TBD" produces exactly that — every rule unparseable, banner green. That contradicts this PR's own stated rule that absent is not passing, and `ExtractionAutovalidatePanel` already handles the case correctly. There's now a third, neutral state for a report with no decisive verdict.

**2. Rules evaluated on a last-document-wins merge.** This was the worst one. `normalized` is built by `normalized.update(entity)` over one entity per document, and every entity carries the same keys — so on a 5-proposal run it *is* document 5's values. That single report was rendered once, above the per-document result selector, and didn't change when the user switched sets.

> Document 1's budget genuinely doesn't add up. Document 5's does. The user looking at Result 1 sees a green banner earned by document 5.

Reports are now produced **per result set** and index-aligned with `results`, the same way `sources` already is, and the strip shows the one for the set on screen.

**3. Stale strip across runs.** The open/reset effect cleared `resultSets`/`resultSourceSets`/`resultDocNames` but never the report, so a red "2 of 3 failed" survived a rail click onto a different run — the exact condition the run path already guards against.

**4. Event-loop block.** `_evaluate_cross_field_rules` was a synchronous call from an `async def`. A `custom_expression` rule goes through `execute_sandboxed_code`, which joins a worker thread for up to 5s plus stop-grace. Behind the opt-in Validate button that stalled one request; on the hot path of every extraction it stalls the whole server, repeatedly, for one bad expression. Now on `asyncio.to_thread`.

While fixing #2 I introduced a cost bug and caught it before pushing: per-entity **plus** the merged report is N+1 sandbox evaluations, so a 20-document run with a custom expression would burn ~a minute of worker time for a report nothing reads. A single result set now reuses its one evaluation.

4131 passed, `tsc --noEmit` clean.

### Not fixed here

- **The rail's restore path** reads only `result_snapshot.normalized` and `.sources`, so reopening a completed run still doesn't restore the rule outcomes. The snapshot now carries `cross_field_sets` and the panel reads it, but `ActivityRail.openExtraction(...)` doesn't pass it through — a small follow-up in the rail's signature.
- **`CrossFieldRunStrip` duplicates `CrossFieldViolationsPanel`**, which already renders this data with `describeRule()` headlines and the "False alarm" escape hatch. The strip shows the raw engine message instead (Python list repr in a red alert) and drops the false-alarm button — which matters more here, since this is now the surface most users hit a false positive on. Worth reusing the panel rather than keeping two.

PR #766, opened 2026-08-28
fix(extraction): badge what the quote supports, not that a quote exists

ViaJables:
Part 1 of 10 from the production-credibility review.

## The bug

The source badge keyed off `verified`. That field records only that a supporting passage was **found somewhere in the document** — not that the passage backs the value the model wrote down.

So an extracted `$4,200,000` that never appears anywhere, paired with a real sentence from the budget narrative, carried the same confident blue `p. 12` chip as a value the quote actually states. The badge was reporting *"a quote exists"* while every reader took it to mean *"this value is supported"*.

Whether the quote supports the value is a separate measurement (`value_supported`), and it was already being computed. The badge simply wasn't reading it.

## What changed

`resolve_entity_sources` now records a `support` state alongside `verified` / `value_supported`, derived by `support_state` with four states:

| state | badge |
|---|---|
| `value_supported` | blue `p. 12` — the quote states the value |
| `quote_unsupported` | grey `p. 12 · unconfirmed` — see below |
| `unassessed` | grey `p. 12 · unconfirmed` |
| `no_source` | no chip |

The frontend's `fieldSupportState` mirrors the derivation and falls back to deriving it when a stored `support` value isn't recognised.

## What this does *not* ship, deliberately

**`quote_unsupported` presents as neutral `unconfirmed`, not as a red alarm.** `ALARM_ON_UNSUPPORTED_QUOTE` is `false`.

`value_supported_by_quote`'s false-positive rate is unmeasured. A red *"quote doesn't match"* on a correct value is the fastest way to teach a grants officer to ignore the badge altogether — which costs more than the over-trust it was meant to fix. So the state is **recorded distinctly and presented neutrally**, and recording it is exactly what makes measuring the rate possible. Flip the constant in Phase 3, with numbers behind it.

The `SUPPORT_BADGES.quote_unsupported` entry and the alarm branch of the toast are present but unreachable while the flag is false.

## Back-compat: historical runs will look different

Every `_field_sources` sidecar written before `value_supported` was recorded carries `verified: true` and nothing else. `support_state_of` never promotes those to supported, so **reopening an older run shows grey `p. 12 · unconfirmed` where it used to show a blue `p. 12`.**

That is the intended rule — an unmeasured quote should not claim support — but it does mean the entire existing run history changes appearance. No migration is possible without re-resolving sources against the documents.

## Scope

The extraction panel is the only surface that renders this badge. `WorkflowEditorPanel`'s per-field table is the unrelated Form Filler report, and `chat_tools.py` does not exist on `main`. Nothing else to update.

## Tests

`test_source_support_state.py` + `test_extraction_sources.py` — 67 passed.

## Note for v5

`RELEASE_NOTES_5_0.md` **cannot yet** claim the product tells you when a quote doesn't back a value. It records that state; it does not yet show it. That claim belongs with the Phase 3 flip.


ViaJables:
## Review round 1 — this skipped a phase gate, and the gate was mine

`/code-review 766 high` flagged the central change, and it's right. `value_supported_by_quote`'s own docstring says the signal is *"recorded but not yet surfaced"* and that *"the distribution can be measured before any of this drives a badge"*. The project note on this work says PR #713 added it **"deliberately not moving the badge until the distribution is measured (Phase 2)"**, with the badge redesign as Phase 3. This PR went straight to Phase 3.

The docstring even names the cases the check gets wrong. Concretely, these now render **red "quote doesn't match"** on correct values:

- `Jane Smith, Department of Chemistry` extracted from *"Dr. Jane Smith of the Department of Chemistry will serve as PI"* — multi-part value, no literal match
- `1200000` against a quote saying *"$1.2 million"*
- any Yes/No or judgement value on a field with no `enum_values` — which is most fields, since it defaults to `[]`

A grants officer shown a fabrication warning on a correct value learns to ignore the badge, and takes the honest states down with it. That costs more than the over-trust this PR set out to fix.

### What changed in aead3e6b

The four states are **unchanged in storage** — that distinction is exactly what makes the Phase 2 measurement possible. Only presentation moves: `quote_unsupported` renders as the neutral "unconfirmed" behind one named flag, `ALARM_ON_UNSUPPORTED_QUOTE = false`, to flip in Phase 3 with numbers behind it. The shared copy no longer asserts *which* unconfirmed case it is, since both mean the same thing to a reader.

The PR still does the thing it exists to do: a blue chip no longer means "a quote exists somewhere".

Also fixed:
- `fieldSupportState` returned `src.support` unvalidated. It arrives from stored snapshots cast `as ExtractionSourceMap`, so nothing checks it at runtime — an unrecognised value rendered **no badge at all**, the silent over-trust this change removes. Now falls back to deriving. 4 tests.
- The click toast derives from the same state the badge shows; a `verified: true, quote: null` sidecar used to toast "No source found" while the badge disagreed.
- The `supported` tooltip says an automatic check *found* the value rather than implying confirmation — it tests whether the value appears in the passage, not whether the passage assigns it to that field.

4121 backend passed, 14 frontend, `tsc --noEmit` clean.

### Follow-up, not fixed here

`support_state_of` (backend) and `fieldSupportState` (frontend) are two copies of the same legacy derivation, and only the frontend one actually runs — the backend copy has no caller in `app/`. The recorded rule for this area is *fix badge logic backend-side, it's the only layer that knows which provenance case it's in*. Reconciling that needs the stored-snapshot read path to backfill `support`, which is its own change.

PR #767, opened 2026-08-28
feat(workflow): workflow and automation extractions carry their sources

ViaJables:
Part 4 of 10 from the production-credibility review. **Extraction trustworthiness — the batch layer.**

`workflow_engine.data_extraction_model` called `engine.extract` with neither `capture_sources` nor `doc_metadata`:

```python
output = engine.extract(
    extract_keys=keys, model=model, full_text=full_text,
    doc_texts=doc_texts, field_metadata=field_metadata,
)
```

The interactive run (`routers/extractions.py:1000`) passes `capture_sources=True` unconditionally. The workflow path passed nothing. So every value a workflow step produced arrived with no quote, no page, and no verification — including the overnight folder-watch automation, which is the highest-volume and least-supervised path in the product and the one a user is least able to spot-check. "Source-linked answers" was an interactive-UI feature the batch layer did not have.

## Alignment is the whole game

The engine pairs `doc_metadata` with `doc_texts` **by position**. A metadata list that drifts by one entry doesn't fail — it silently attributes every quote to the wrong document and reports a page from the wrong file, which is worse than reporting none.

So the texts and their metadata are now built in a single pass, `_build_extraction_inputs`, returning `(text, meta)` pairs. `_build_extraction_texts` stays as the texts half of it, so nothing that called it had to change. The alignment cases that used to be possible to get wrong:

- **A document with no extracted text** is skipped in the texts list — so its metadata is skipped with it, rather than shifting every later document by one.
- **A previous step's output** has no document behind it, but still occupies a slot with an empty metadata entry. Its quote is verified against that text and simply resolves to no page — traceable to the step, honest about having no page.
- **Mixed sources** (`step_input` + `workflow_documents`) keep both of the above true at once.

## The metadata had to be hydrated first

`doc_metas` and `selected_doc_meta` were attached only when `task_doc["name"] == "FormFiller"` — in all three hydration sites (full run, Test Step, and the passive automation path). An Extraction node never saw a `text_markers` array, so even had it asked for sources it could not have resolved a page.

The gate is now `DOC_META_TASKS = ("FormFiller", "Extraction")`, in `form_fill.py` beside `document_meta` and imported by both task modules. Still a gate rather than unconditional: attaching markers to nodes that never read them is just payload.

## One display fix

`format_extraction_results` iterates `item.items()`, so the new `_field_sources` sidecar would have rendered as a bullet containing a JSON dump of every quote — in the step's markdown, which is often the workflow's actual deliverable. It's skipped.

## Tests

`test_workflow_extraction_sources.py` — 10 cases, weighted toward alignment: blank-document skip, mixed sources, missing `doc_metas`, selected-document metadata, forwarding through `data_extraction_model` to the engine, the sidecar staying out of the formatted output, and `_build_steps_data` actually attaching `doc_metas` to an Extraction task. 349 passed across the workflow suites.

## Branch choice

`workflow_engine.py` diverges from `major/agentic-chat` by 4 lines, all inside `KnowledgeBaseQueryNode` (a `DocumentManager` singleton refactor); `data_extraction_model` is identical. `form_fill.py`, `workflow_tasks.py` and `passive_tasks.py` are byte-identical in the regions touched. This lands on `main`.


ViaJables:
## Review round 1 — the sidecar had to travel *beside* the output

`/code-review 767 high` confirmed the alignment work is correct (it traced `_build_extraction_inputs` against all three hydration sites and `_resolve_sources`), but found that this is the only `capture_sources=True` caller that leaves `_field_sources` **on** the entity dicts. Every other one pops it — `routers/extractions.py:1021`, `chat_tools.py:1470` and `:2558`, `chat_service.py:838` — and `extraction_tasks.normalize_results` special-cases skipping it.

That shape is load-bearing in three places:

**`DataExportNode` raises mid-export.** Headers come from `input_data[0].keys()` and `csv.DictWriter` defaults to `extrasaction='raise'`. The engine attaches the sidecar only `if sidecar:`, so a multi-document run where document 1's reply carried no quotes and document 2's did produces rows with different keys → `ValueError: dict contains fields not in fieldnames: '_field_sources'`, surfacing as a **failed run** on what is often the deliverable. When row 0 *does* have it, the export instead gains a column of Python dict reprs. There's a test for exactly that mixed shape now.

**Approval artifacts degrade.** `detect_artifact_kind` classifies an extraction result as `extraction_table` only when every value is `str|int|float|None`. A dict value drops it to `ARTIFACT_JSON`, so an Extraction → Approval workflow hands the reviewer a raw JSON blob instead of the editable field table — with the provenance exposed and hand-editable.

**Downstream prompts inflate.** `_build_combined_context` returns the raw payload for a single source and `llm_chat_model` json-dumps it into the `CONTEXT:` block under "answer using ONLY the CONTEXT" — so quotes, pages, document ids and support flags ride into every downstream Prompt/Formatter step, frequently several times the size of the values.

Now carried as `field_sources` beside the output, like Form Filler's `fill_report`, with `MultiTaskNode` passing it through the wrapper.

### Two more from the same review

**A `step_input` quote was recorded as verified with no document.** `resolve_entity_sources` sets `verified = offset is not None` — "the quote was located in the text we searched" — and for that slot the text is a previous LLM step's own output. A quote found in the model's own words was indistinguishable from one located in a source document except by a null `document_uuid`. The slot is tagged `kind: "step_input"` now, as Form Filler's equivalent already was.

**The passive path never hydrated `selected_doc_text`.** The other two sites do; automation didn't, so an Extraction step configured with "Selected Document" ran under folder-watch with **no text at all** — an empty answer produced nightly on schedule, with nothing saying the input was missing. Pre-existing, but it sat in the block this PR edited and undercut its parity claim.

4127 passed. Fixed in 5577de89.

### Follow-up

The review also noted the sidecar inflating `workflow_result.steps_output` in Mongo, which feeds `workflow_service`'s validation judge prompt through a `[:20_000]` truncation — so a multi-step run's later steps could be truncated out of "Intermediate Step Outputs" and silently change quality verdicts. Splitting the sidecar out shrinks that a lot, but the truncation is still a sharp edge worth revisiting on its own.

PR #768, opened 2026-08-28
fix(validation): a judge outage is not a score of zero

ViaJables:
Part 5 of 10 from the production-credibility review. **Validation honesty.**

`judge_field_value` swallowed every exception into a verdict:

```python
except Exception as e:
    return {"score": 0.0, "verdict": "FAIL", "reasoning": f"judge error: {e}", ...}
```

Swallowing is right — one judge error must not crash a trial loop. Scoring it 0.0 FAIL is not. "The judge is down" and "the extraction was wrong" are different facts, and the first was being recorded as the second: during a provider outage every field scores zero, the trial records a catastrophic accuracy, and that number goes into the quality history **permanently**, where the next comparison reads it as a regression and alerts on a problem that never existed.

## The verdict

`JUDGE_UNAVAILABLE` with `score=None`, and `is_judge_unavailable()` as the check. Not a new score value — a different kind of answer, so no caller can average it in by accident. A `None` score raises where a `0.0` would have silently propagated, which is the point.

## Aggregates exclude it

- **`judge_test_case_extraction`** averages over fields actually *scored*. Two fields at 1.0 and one unreachable is 1.0, not 0.67. It now also reports `num_fields_unavailable` and `judge_coverage`.
- **`extraction_tuning_service`** no longer writes `judge_scores[key] = 0.0` on failure (in either the returned-verdict path or the raised-exception path). A field's accuracy averages only the runs that were judged; a field the judge could not reach on *any* run goes unmeasured for that trial rather than contributing a fabricated zero. `total_comparisons` — the N behind the standard error — counts scored comparisons, not attempted ones.

## A coverage floor, because excluding is not enough

Excluding unavailable fields keeps the arithmetic honest but makes a number computed from three of forty fields look exactly like one computed from forty. So `MIN_JUDGE_COVERAGE = 0.8`, applied at every point where a tuning result becomes a stored number:

| site | before | after |
|---|---|---|
| trial | scored, could win | `status: judge_unavailable`, score/accuracy/lift nulled, excluded from winner selection (which already filters to `completed` with a score) |
| `baseline-no-tool`, `baseline-default` | stored | withheld — a baseline poisons every comparison measured against it |
| `holdout-winner`, `holdout-default` | became the headline `optimized_score` | withheld |

The floor is deliberately high: discarding a trial costs one wasted config, keeping a degraded one costs a permanent wrong number in the quality history.

## And σ stops inflating

The two `_rejudge` closures fed `float(v["score"])` into `sample_judge_variance`. A failed re-judge returned 0.0, contributing a delta of `-original` and inflating σ — which widens the significance gate, so the optimizer declines real improvements on evidence that does not exist. They now raise, and `sample_judge_variance` already drops a raising sample.

## Tests

`test_judge_unavailable.py` — 9 cases: the outage verdict itself, `is_judge_unavailable` discrimination, the exclude-don't-zero aggregate, full coverage, the trial gate (uncovered / covered / strict-match-never-gated / floor boundary), and a withheld baseline. The existing `test_judge_field_value_swallows_exceptions_returns_fail` is updated to the new contract and renamed.

## Branch choice

`extraction_judge.py` is byte-identical between `main` and `major/agentic-chat`; `extraction_optimizer.py` and `extraction_tuning_service.py` diverge nowhere near these functions. Lands on `main`.


ViaJables:
## Review round 1 — the gate could not see the outage, and the apply path got *less* safe

`/code-review 768 high` found seven issues. The two that mattered:

**1. The coverage gate was blind to a total outage.** `judge_coverage`'s denominator counted deterministic pre-judge resolutions. Those never call a model and cannot be unavailable, so on a set where 100 of 120 comparisons short-circuit, a **complete** LLM-judge failure gave coverage `0.833` — clear of the 0.8 floor. The gate passed; every ambiguous field silently dropped out of the score. Coverage is now measured over comparisons that actually needed the LLM. Three new tests pin the arithmetic, including the fully-deterministic case (1.0, not a division by zero).

**2. Withholding the baseline removed the significance gate entirely.** This is the one I'd flag hardest:

```python
tied_with_baseline = baseline_default_score is not None and abs(top - baseline) <= band
...
if apply_on_finish and run_doc.best_config and not run_doc.tied_with_baseline:
    await _apply_best(...)   # writes to the user's live extraction config
```

`tied_with_baseline` is False both when the winner genuinely beat the baseline **and when there is no baseline at all** — and withholding an uncovered baseline made the second case reachable. So a judge outage silently converted the significance gate into no gate, and applied a config to live settings on no evidence. My safety change removed a safety check. Applying now requires a baseline to measure against.

Also fixed: the best-so-far ticker read the raw score (a gated trial still set "Best so far" and reset the convergence counter); a withheld holdout left the headline at the in-sample score this code documents as best-of-N inflated, unflagged; the apply preview gave dropped fields `baseline: 0.0`, reporting a fabricated ~100pp improvement in the modal that gates the apply; and both trial renderers drew a null score as a red **0%**, putting the outage back on screen as the quality collapse this PR exists to stop reporting.

`avg_score` returns None rather than 0.0 when nothing was scored — "nothing to judge" is not "everything failed". The existing empty-keys test asserted the old contract and is updated.

4126 passed, `tsc --noEmit` clean. Fixed in 6e64759b.

PR #769, opened 2026-08-28
feat(quality): a detected regression reaches a human and changes the badge

ViaJables:
Part 6 of 10 from the production-credibility review. **Validation loop closure.**

`quality_monitor` detects a regression, and then:

```python
await QualityAlert(alert_type="regression", ..., severity="critical" if delta >= 20 else "warning").insert()
# Phase 6: auto-enqueue a shadow optimizer...
```

`QualityAlert` is rendered in exactly one place — the **admin** Quality tab, behind `_require_admin`. The shadow optimizer run lands in the optimizer inbox. And the item keeps advertising the tier it held before the drop; the code comment says it outright, it "Does NOT un-verify."

So the loop closed on the system and never on a person. The user who owns the item — the one who will open it again tomorrow and act on what it produces — was told nothing, and every surface kept endorsing something monitoring had already decided was broken. That is worse than not monitoring: it manufactures a green badge the system's own data contradicts.

## The badge stops claiming what the system no longer believes

`VerifiedItemMetadata` gains `regression_pending_review` (plus when, how bad, and the score to get back to). `QualityBadge` renders it as a red **"Regression pending review"** that *replaces* the tier rather than decorating it — a dimmed "Good (72%)" still reads as an endorsement, and the point is that the previous rating no longer applies. Hovering says why.

It travels through every surface that shows quality: the extraction and workflow editor panels, the library rows, and the verified catalog — the catalog especially, because that is where a user who knows nothing about an item decides whether to trust it.

## Somebody gets told

- **Every regression** notifies the item's owner in the bell — resolved from the underlying `SearchSet` / `Workflow` / `KnowledgeBase`, deep-linked to the item with the same `/?extraction=`, `/?workflow=`, `/?kb=` links `failure_notifications` already uses, and coalesced per item so a nightly monitor that keeps finding the same regression keeps it visible without burying everything else.
- **A `critical` regression also emails them.** A bell entry is the right weight for a warning; a 20-point drop on something the owner may be about to rely on has to reach them where they actually are.

Best-effort throughout, in the shape the failure-notification rules already set: a missing owner, an absent email address, or a notification failure logs and moves on — the flag is still set, and one bad item can't take down a nightly pass.

## Coming back out of it

Two ways, and neither needs a ceremony:

- **Automatically**, when a later validation recovers the pre-regression score. A transient dip shouldn't need a human. The recovery bar only ever *rises* — a second, smaller drop while a review is pending must not make recovery easier than the first drop did.
- **By acknowledging the alert**, which is the review the item was waiting for. Without this the flag would be permanent and the signal would stop meaning anything.

## One finding that did not become code

The review also asked to "auto-revalidate templates with test cases regardless of verified status." Checking it: `update_quality_metadata` creates a `VerifiedItemMetadata` row for **any** item that has ever been validated, verified or not, and the revalidation query filters on `last_validated_at` with no verification-status condition — so unverified items already are revalidated. The real remaining gap is narrower: an item validated *zero* times has no row and is invisible to the monitor. Closing that means validating every search set in the system nightly, which is a token-cost decision rather than a correctness fix, so it is flagged here rather than shipped.

## Tests

`test_quality_regression_review.py` — 12 cases: warning flags + notifies without email, critical emails the owner, a missing owner or a failing notification still leaves the flag set, the recovery bar only rises, automatic clearing on full recovery, partial recovery leaves the flag up, `clear_regression_review` idempotency, and owner resolution per item kind including a malformed workflow id. `QualityBadge.test.tsx` gains 2. Full backend suite green; `tsc --noEmit` clean.

`_make_verified_metadata` in `test_quality_service.py` now sets the new fields explicitly — a bare `MagicMock` attribute is truthy, so without it every fixture would have looked like it had a regression pending review.

## Branch choice

`quality_tasks.py` is byte-identical between `main` and `major/agentic-chat`; the other touched files diverge nowhere near these blocks. Lands on `main`.

**Note for v5:** `docs/QUALITY_SIGNALS_EXPLAINED.md` (branch-only) publishes Good = 75–89 while `DEFAULT_QUALITY_CONFIG` uses `good: min_score 70`. That correction is a branch edit regardless of where this lands, and belongs in the same pass so the two don't drift further.


ViaJables:
## Review round 1 — the badge never rendered, and every test passed

`/code-review 769 high` verified the service logic is sound (it checked the `create_notification` and `send_email` signatures, `Settings.frontend_url`, the deep-link router params, the owner fields on all three models, and that the `{count}` group title can't blow up on a user-supplied item name). Four findings, all fixed in 745d4ef4.

**1. The headline behaviour didn't work.** `_attach_quality` emits `regression_pending_review`; `SearchSetResponse` never declared it; Pydantic's default `extra="ignore"` drops undeclared keys on the way out. The extraction header badge got `undefined` and **never showed a regression** — on the surface the feature exists for.

What makes this worth dwelling on is that nothing caught it. The service-layer tests exercise the flag directly and pass. The Quality Pulse card kept working because it reads a *separate* endpoint returning a plain dict, so manual testing would have looked fine too. Every signal I had said the feature worked. There's a test on the response model now, not just the service.

**2. The catalog grid tile was missed.** Only the detail modal got `regressionPending`; the tile still showed the old tier badge, gold shield and tier ring — the surface this PR's own description singles out as mattering most, since it's where someone who knows nothing about an item decides whether to trust it.

**3. Acknowledging *any* regression alert cleared the flag.** Regression alerts have no unacknowledged-dedup (unlike `stale` / `config_changed`), so an item can carry several. Acknowledging an older one un-flagged an item a newer, worse, still-open alert covered — and destroyed the recovery baseline with it, since `_reset_regression_state` clears `regression_baseline_score` too. The flag now lifts only when no regression alert for that item is still open.

**4. The bell had no icon or colour for `quality_regression`,** and the kind isn't in `FAILURE_KINDS` — so a warning-severity regression (the ones that get no email by design) rendered as a generic grey bell, indistinguishable from an informational notice.

4129 passed, `tsc --noEmit` clean.

---

That closes the last outstanding review round in this batch — all ten PRs (#764–#773) have now been reviewed at `high` and remediated.

PR #770, opened 2026-08-28
fix(ingestion): detect and disclose a partial extraction

ViaJables:
Part 7 of 10 from the production-credibility review. **Ingestion honesty.**

An extraction can succeed and still not be the whole document. Two ways it could, and neither was visible anywhere.

## 1. A partial OCR conversion was a log line

`parse_docling_response` keeps `partial_success` text — correctly, partial text beats falling all the way back to PyMuPDF — but announced it only to the log:

```python
if errors:
    logger.warning("Docling conversion reported errors but returned %s: %s", field, ...)
return content
```

So a 400-page package whose converter gave up after thirty pages was stored as a **complete** document. Every answer drawn from it was confidently about a fraction of it, and nothing in the answer, the document row, or the extraction result would have said so.

It now fills in an optional `report` dict, threaded through `convert` (both the sync and async-poll paths) → `ocr_extract_text_from_pdf` → `extract_text_with_markers` → the ingestion task, which stores `partial_ocr` on the document. Optional keyword throughout, so no existing caller changes. If the reader ends up falling back to PyMuPDF, the flag is cleared — the text we return is no longer the partial text — and restored if the fallback fails and we ship the OCR text after all.

## 2. `MIN_PDF_TEXT_LENGTH` cannot see a sparse extraction

100 characters is a **whole-document** floor. 150 characters is an ordinary result for a cover page and a total failure spread over 400 pages, and both clear it identically. `is_sparse_extraction` adds the per-page equivalent: `MIN_CHARS_PER_PAGE = 40`, set well below any plausible real page so crossing it means the OCR gave up rather than that the document is terse. Whitespace does not count toward the density, since a stalled conversion often returns page scaffolding and little else.

## 3. `provider="raw"` accepted an error page as document text

```python
if resp.status_code != 200:
    raise ...
return resp.text
```

No content-type check. A wrapper answering HTTP 200 with an HTML error page had that page become the document's text — chunked, embedded, and answering questions. Now the body must look like extracted text: an unacceptable `Content-Type` is refused, and so is a body opening with markup even when the header claims `text/plain`. A **missing** `Content-Type` is not itself grounds for rejection — plenty of small OCR wrappers send none — but the body check still applies. It raises `OcrRequestError`, so the existing retry-then-fall-back-to-PyMuPDF path handles it like any other failed attempt.

## Surfacing, the way `extraction_low_quality` already is

`ingestion_warnings` on `SmartDocument` holds the codes; `document_service` owns the labels so every surface says the same thing:

- **File list and poll status** carry the codes alongside `extraction_low_quality`. Reprocessing clears them, in every branch that already cleared the nonletter ratio — otherwise a fixed document keeps a warning from a previous run.
- **Chat** gets its own notice, deliberately separate from the garbled-text one. "The text layer is unreadable" and "this is only part of the document" call for different advice, and a confident summary of pages 1–30 of a 400-page package is the failure mode that reads as a perfectly good answer.
- **Extraction** discloses degraded inputs for the first time. `run_extraction_sync` collects them into an optional out-list (leaving the historical `list[entity]` return alone) and `/run-sync` returns `document_warnings`. Chat had this warning and extraction did not, which is backwards: a wrong extracted deadline or dollar amount is the more expensive place to stay quiet.

## Tests

`test_partial_ingestion.py` — 22 cases: partial-conversion reporting (including errors-without-partial-status, a clean conversion reporting nothing, and the report staying optional), the raw-body checks (HTML page, lying content type, plain text, absent header), the density floor (400 pages × 150 chars, the same 150 chars on one page, whitespace padding, unknown page count, the boundary), and the surfacing helpers (readable text, retired codes ignored, legacy documents with no field, chat's title list).

## Merge note

`/run-sync`'s response dict is also touched by **#765** (cross-field rules on every run), which is stacked behind #764. Whichever lands second takes a two-line conflict in that return.

## Branch choice

`ocr_client.py`, `document_readers.py`, `document_tasks.py` and `document_service.py` are byte-identical between `main` and `major/agentic-chat`; `search_set_service.py` diverges only at 403–520, above `run_extraction_sync`. Lands on `main`.


ViaJables:
## Review round 1 — the content-type gate accepted exactly what it rejected

`/code-review 770 high` found four issues. The headline one is the most embarrassing in this batch:

```python
_ACCEPTED_RAW_CONTENT_TYPES = ("text/plain", "text/markdown", "application/json", "text/")
if content_type and not content_type.startswith(_ACCEPTED_RAW_CONTENT_TYPES):
```

`"text/html".startswith("text/")` is `True`. The gate written to reject HTML error pages **accepted them** — it was a no-op, and only the body sniff was doing any work. My own HTML test passed *via that sniff*, so it never noticed the mechanism it was supposed to prove was inert. It's an exact-match set now, and the sniff looks past a BOM and any leading comment, since `<!-- generated by gateway -->` in front of an error page slipped through.

**The rejection was also classed retryable.** `status_code=200` isn't in `PERMANENT_STATUS_CODES`, so a wrapper that *always* answers 200 with an error page burned all three in-process attempts, raised `OcrUnavailableError`, and drove ~25 minutes of Celery backoff before the document was written off — instead of the PyMuPDF fallback this PR and the task's own comment both promise. A misconfigured wrapper is not an overloaded one, and waiting doesn't fix it. `OcrRequestError` takes an explicit `permanent` flag rather than overloading the status code, since 200 is not inherently permanent.

Two smaller ones:

- `low_quality_text` was emitted as a warning code by the extraction path but had no entry in `INGESTION_WARNING_LABELS`, so `ingestion_warnings()` filtered it right back out — the run reported a warning that rendered as nothing.
- `duplicate_project` copied `raw_text` and the nonletter ratio but not `ingestion_warnings`, so duplicating a project silently upgraded a partially-ingested document to a clean one. Guarded with `getattr` to match how `document_service` already reads the field, which also covers documents stored before it existed.

4139 passed. Fixed in fcd6c587.

### Still open — the review is right and I haven't fixed it

`document_warnings` and the new `ingestion_warnings` on the file list / poll status have **zero references in `frontend/src`**. Of the four surfaces the PR body claims, only chat actually renders anything today. The extraction disclosure — the PR's headline change — is invisible without a frontend change, and `document_warnings` is also left out of the run's `result_snapshot`, so a reopened run wouldn't show it either.

I'd rather flag that than quietly ship a PR whose description overstates what a user sees. It wants its own change: a warning strip in the extraction panel plus the snapshot write.

PR #771, opened 2026-08-28
fix(workflow): a retried run resumes instead of re-executing every step

ViaJables:
Part 8 of 10 from the production-credibility review. **Operational honesty.**

```python
@celery_app.task(
    autoretry_for=TRANSIENT_EXCEPTIONS,
    max_retries=3,
    ...
)
def execute_workflow_task(self, workflow_result_id, ...):
```

The generic handler re-raises into that decorator, and the task had no resume index. So a provider read timeout on step 4 restarted the workflow **from step 0**, up to three more times. An `APICallNode` POST on step 2 fired four times. A `save_to_folder` on step 2 wrote four copies. Every earlier step's tokens were billed four times over. And none of it was visible — each attempt looked like a normal run.

Retrying is the right response to a transient failure; restarting is not. The engine has supported resuming since the approval gate was built — `execute(start_index=..., initial_output=...)`, with `_replay_step_entries` to stitch the record back together — so a retry now uses the same mechanism the approval path already proves.

## Where it picks up

`num_steps_completed` is the index of the step that *finished*, and its output is already in `steps_output`, so the retry starts one past it and is fed that output as its input. `_resume_point` makes that decision and returns `(0, None)` — a full rerun — whenever the persisted state can't justify skipping anything:

- no completed steps, or an absent count
- **the output for the last completed step is missing or isn't a dict** — resuming would feed the next step nothing and produce a confidently empty run, which is worse than repeating work
- **more completed steps than the engine has nodes** — the workflow was edited between attempts, and replaying old outputs against a re-shaped graph would attribute them to the wrong steps

Every one of those falls back to the old behavior. The change only ever *skips* work it can prove was done.

## Two supporting corrections

The "update result to running" block unconditionally wrote `num_steps_completed: 0, steps_output: {}` — which is what erased the progress a retry needed. A resuming retry now keeps both; a fresh attempt still clears them.

And `execute()` reports only the steps *this* pass ran, so a resumed pass's `data` is prefixed with the earlier steps replayed from `steps_output`. Without it the saved run record would begin mid-workflow and everything before the failure would vanish from the output — the same correction the approval resume already makes.

Resuming is gated on `self.request.retries`, so a first attempt is byte-for-byte the behavior it has today.

## Tests

`test_workflow_resume_on_retry.py` — 7 cases on `_resume_point`, weighted toward the refusals, because a wrong resume point is the failure nobody notices until a side effect has fired twice: normal resume, no completed steps, missing output, non-dict output, an edited workflow, the last step as resume point, and an absent count. 100 passed across the workflow task/retry/approval suites; full backend suite green.

## Branch choice

`workflow_tasks.py` diverges from `major/agentic-chat` by 14 lines at 1147/1168 (result-saving), well clear of the retry decorator and the execute block. Lands on `main`.


ViaJables:
## Review round 1 — I fixed one path and left the identical bug in the one below it

`/code-review 771 high` verified the resume arithmetic against the engine (the `num_steps_completed: idx` / `steps_output[keys[idx]]` pairing, the `len(keys)` vs `num_steps_total` alignment, and that `_mark_workflow_failed` doesn't clobber the state a resume depends on). Three findings, all fixed in ec9d94fa.

**1. `resume_workflow_after_approval` has the same bug, untouched.** Same `autoretry_for` + `max_retries=3`, no resume index of its own — it passed `start_index=step_index + 1` on *every* retry, so a transient failure at step 8 re-ran steps 4–7 up to four times. Same bug class, one function down from the one I fixed. It now takes whichever is further along, the gate or the persisted progress, since a first pass through that task must still start at the gate.

**2. The finalize tail could still double-write.** This is the sharper one, because it's the exact harm this PR's description advertises fixing, reachable by a different route:

> The tail sits **after** `engine.execute()` and **outside the try**. An `AutoReconnect` on the `num_executions` `$inc`, or a Redis blip on `auto_validate_workflow.delay`, retries the task. The resume then correctly skips every step — and lands right back on `save_results_to_folder`. The filename template carries `{time}`, so that's a **new** library document, not an overwrite. Up to four copies.

Claimed once now via an atomic `finalized_at` on `WorkflowResult`. `{"finalized_at": None}` matches a missing field in MongoDB, so runs written before this stay claimable exactly once rather than never.

**3. An ordering regression I introduced in the first commit.** The `status`/`num_steps_total`/`output_step_names` write moved *below* `build_workflow_engine`, which is outside the try — so `ValueError("Unknown task type: ...")` left the row at `queued` with empty `output_step_names` instead of `running`. Both states are stuck, but `queued` hides a run that's never coming back. The status write is back above the build; only the progress fields wait for the resume decision.

4123 passed.

One thing the review checked and cleared, worth recording: the resume feeds the next step its predecessor's *persisted* output, which has been through `_bson_safe` — that coerces tuples/sets to lists and bytes to str, but every binary-carrying node (`DocumentRendererNode`, `FormFillerNode`, `PackageBuilderNode`) base64-encodes to an ASCII `str` before it reaches the output dict, so the round-trip is lossless in practice.

PR #772, opened 2026-08-28
fix(validation): give the drift ledger a life and scope its κ to the model

ViaJables:
Part 9 of 10 from the production-credibility review. **Validation stack vs. its marketing.**

The κ-drift ledger was structurally incapable of ever firing, and the κ it publishes belongs to a model most deployments do not run.

## The ledger could not accumulate

`judge_drift.assert_no_regression` returns silently when `trailing_median` has fewer than 3 entries. `backend/tests/fixtures/judge_drift_history.jsonl` is **0 bytes**. The only writer is `test_tier3_llm.py:317`, which runs weekly inside a GitHub Actions runner — with **no step that commits the file back**.

So every run appended one line to an ephemeral disk and threw it away. Drift detection has never had a baseline and, as written, never would. The module's own docstring says "the file is checked in and appended to in CI"; this makes that true:

- A commit-back step in `integration-llm.yaml`, with `contents: write` (the default token is read-only), rebasing before push so a concurrent landing doesn't fail the job, and `[skip ci]` on the commit so it doesn't trigger a build.
- `if: always()`, deliberately — a run that tripped the absolute κ floor is exactly the history a later median wants, and dropping it would bias the baseline upward.

## The κ belongs to the model it was measured on

`judge_contract` publishes `min_kappa = 0.65` for the extraction judge, measured against `INTEGRATION_LLM_MODEL` in this project's CI. The judge model a deployment actually uses is the user's current model (`extraction_optimizer.py:285`), which on a self-hosted install is whatever they configured — commonly a local 8B. That deployment inherits an agreement number established against a frontier model in someone else's CI, which is precisely the substitution a published κ exists to rule out.

`calibration_for(surface, judge_model)` returns a figure **only** for the exact model it was measured on, and `None` otherwise. None means one thing — nobody has ever measured this model on this surface — and it is never filled in from another model's row. That default is the design: the honest answer is what a caller gets without having to remember to ask for it.

`calibration_status` builds on it for a whole deployment: each configured model reported as calibrated or not *for itself*, plus the context a reader needs — the published floor, which models have ever been measured, how many ledger entries exist, and `drift_detectable`, which is false below three entries because the regression check cannot fire there however far agreement moves.

## Where it shows up

A **Judge Calibration** panel in the admin Quality tab, since the κ number was previously visible nowhere in the app at all — only in docs. Each model is a chip reading either `κ 0.81` or **`κ unmeasured`**, and a surface whose ledger is too thin carries a "drift undetectable — N of 3 ledger entries" badge. The intro says plainly that the published floors were measured against this project's CI model and not against yours.

The panel is fetched independently and failure-tolerantly, like the config call beside it: an error there must not blank the tab's real payload.

## Tests

`test_judge_calibration_scope.py` — 13 cases. The ones that matter most are the refusals: an unmeasured model gets nothing rather than someone else's number, a measurement on another *surface* doesn't count, and an empty ledger measures nothing. Plus the latest-of-several selection with a run count, per-model status across a mixed deployment, the two-vs-three-entry drift-detectability boundary (checked against `trailing_median` returning None), `measured_models` reflecting history rather than configuration, and an end-to-end record-three-then-detect-a-regression that demonstrates exactly what the missing commit-back step was denying the ledger.

## Branch choice

`judge_drift.py`, `judge_contract.py` and `integration-llm.yaml` are byte-identical between `main` and `major/agentic-chat`. Lands on `main`.


ViaJables:
## Review round 1 — the fix for "this can never accumulate" could never accumulate

`/code-review 772 high` found two independent reasons this PR didn't do what it says. Both verified against the actual repo config, not read off.

**1. The ledger never reaches a deployment.** `backend/.dockerignore` line 5 is `tests/`, and the Dockerfile copies `app`, `certification-data`, `scripts`, `seeds` — nothing else. `compose.yaml` builds with `context: ./backend`, the supported deploy path. So `/app/tests/fixtures/judge_drift_history.jsonl` doesn't exist in any install, `load_history` returns `[]`, and the panel reports **"κ unmeasured"** for every model and **"drift undetectable — 0 of 3 ledger entries"** for every surface. Unconditionally — including a deployment running exactly the model CI measured. The feature only ever worked in a source checkout.

Un-ignored and copied now. I verified the `.dockerignore` negation actually works with a real `docker build` rather than assuming — it's a known gotcha, and if it had failed the `COPY` would break the build, which is worse than the inertness.

**2. The commit-back could never land.** `main` has classic protection with six required checks:

```
Backend Release Checks, Frontend, Backend Tests (py3.11), Backend Tests (py3.12),
Backend Tier 2 (MongoDB), Backend Tier 4 (ChromaDB)     enforce_admins: false
```

A commit pushed from Actions carries none of them, and `[skip ci]` guarantees they never run → `GH006: Required statuses must pass`. The weekly job would have **failed every run** while the ledger **still never grew**. It opens a PR now — no bypass token needed, and a human on a file that gates a published quality number is arguably the right amount of ceremony. The trade-off is written into the workflow rather than glossed: *the ledger only accumulates when those PRs are merged.*

**3. `if: always()` was decorative.** `record()` was called *after* the floor asserts in `test_tier3_llm.py`, so a run that tripped the κ floor raised before writing anything — `git diff --quiet` succeeded and the step committed nothing. My comment claimed those runs were "exactly the history a later median wants" while the code threw them away. Recording happens first now: the measurement is a fact whether or not it clears the bar.

**4. The gate still pooled judge models** — which is the exact thing this PR set out to scope. Once the ledger accumulates, the first run after `INTEGRATION_LLM_MODEL` rotates would be compared against the **previous model's** median: a spurious regression on the "judge model is silently swapped" case the module docstring names as the reason the ledger exists, then a permanently mixed baseline. `assert_no_regression`/`trailing_median` take a `judge_model`, and tier-3 passes it. 3 new tests, including that a real same-model regression still fires.

**5.** `calibration_for` takes the newest entry by timestamp rather than the last line, since the PR flow can reorder the file on a conflict resolution.

4127 passed, 14 in the calibration suite. Fixed in e593dc1e.

PR #773, opened 2026-08-29
fix(chat): the agentic KB path hedges pages like the classic one does

ViaJables:
Part 10 of 10 from the production-credibility review. **RAG chat honesty — and the only one of the ten that is genuinely branch-only.**

`chat_service`'s classic KB path routes every citation through `page_locator`. The agentic path — the one `RELEASE_NOTES_5_0.md` calls the product — did this:

```python
"page": page if isinstance(page, int) else None,
```

Reading `metadata["page"]` straight off the chunk, which reintroduces both of the bugs `page_locator` was written to prevent, in the flagship feature:

1. **An interpolated OCR page stated as exact.** OCR returns no page structure, so the backend estimates boundaries by spreading the known page count evenly across the text. On a 400-page package with a 40-page budget table that is off by a large and unbounded amount. `page_approximate` marks it; this path dropped the flag, so every estimate rendered as a measured location.
2. **A chunk cited by the page it starts on.** A passage sitting past a page break inside the chunk was cited one page early — the "p. 2 for text footed Page 3 of 5" support ticket.

## What changed

`search_knowledge_base` now uses `cited_pages` (the page of the segment that shares the most query terms with the question, or the range when that's ambiguous — never the starting page as a guess) and `annotate_chunk_pages` (`[p. N]` markers at each internal break, so the model can see where the page changes instead of inferring it). Results and citations carry `page`, `page_end` and `page_approximate`.

## The citation renderer was already waiting for it

`ChatMessage.tsx` renders citations through `formatPageLocator(c.page, c.page_approximate, c.page_end)` and has since #603. It was simply never being sent the flag, so the tilde and the range appear as soon as the backend emits them — no change to that component.

## But one frontend change *is* required

`ToolCallDisplay.tsx` derives its click-to-highlight phrase from the passage text. Now that passages carry injected `[p. N]` markers, a chunk whose page break falls inside its first ~60 characters would open the viewer on a highlight that appears nowhere in the document and matches nothing.

So the tool result also carries `content_verbatim` — the unannotated text — and `pickHighlightPhrase` reads that instead. It is only set when the two differ, and the UI falls back to `content`.

`content_preview`, which drives the separate *citation* click path in `ChatMessage.tsx`, is built from the unannotated text for the same reason.

## And the model needs the labels, not just the tilde

The classic path prepends two explanations, for a measured reason recorded in its own comment: a tilde the model has not had explained to it gets normalised away and the estimate is restated as fact — five times out of five at temperature 0 on a fully interpolated document. The same two labels now lead the tool result, before the passages they govern:

- what `page_approximate` means, with "never state one as exact and never say a passage is *explicitly* or *clearly* on it"
- what `[p. N]` inside a passage means, with "cite the page the passage you are using falls under"

Both are emitted only when they apply, so an all-measured, non-spanning result is unchanged.

## Tests

`test_agentic_kb_page_hedging.py` — 10 cases: approximate marked / measured not marked / the model told what the estimate means / no note when nothing is estimated; the answering page beating the starting page, an ambiguous match citing the range instead of guessing, break markers in the text the model reads, and a spanning-*and*-approximate chunk hedging its own break markers (`[p. ~3]`); plus sheet-only and location-less sources. 97 passed with `test_chat_tools.py`; full backend suite green.

## Why this one is branch-only

`chat_tools.py` exists only on `major/agentic-chat` — it is the v5 agentic layer. Nothing to port to `main`; `page_locator` and the frontend renderer are already there and unchanged.


ViaJables:
## Review round 1 — both shape changes broke a consumer I didn't check

`/code-review 773 high` confirmed the page-hedging logic itself is right (it verified `cited_pages`/`annotate_chunk_pages` are used the same way `chat_service.py:2841` uses them, and the bool-vs-int guards hold). Both findings are about the two changes this PR makes to the tool's **return shape**.

**1. The prepended note is rendered as a passage.** `ToolCallDisplay` treats every element of a `search_knowledge_base` result as a passage:

- `Found ${content.length} relevant passages` → 5 passages report as **6**
- `passages.slice(0, 3)` → the note takes a preview slot and renders as a blank grey "Source ·" row with no content, so only 2 real previews show
- `toolResultToText` → "Copy passages" leads with an empty `[Source]` block

And it fires on the **common success path** — any scanned KB makes `any_approximate` true. The guidance now rides on each entry as `citation_note`; the array stays homogeneous. (The pre-existing `no_results` return has the same shape problem, but that path is rare and I've left it alone rather than widen scope.)

**2. The `[p. N]` markers leaked into document highlighting.** `entry["content"]` is both what the model reads *and* what the UI uses for click-to-highlight — `actions.setHighlightTerms([pickHighlightPhrase(chunkContent)])`, and `pickHighlightPhrase` takes the first ~60 characters. A chunk whose page break falls inside that window put an injected `[p. 3]` into the search phrase. That string appears nowhere in the document, so clicking the passage opened the viewer on a highlight that **can never match** — on a feature whose entire point is letting someone verify a citation.

The classic path avoids this by keeping annotated text model-only. There's now a `content_verbatim` for anything matching against the document, set only when the two differ, and the UI uses it.

4 new tests. **4429 passed, `tsc --noEmit` clean** — worth stating explicitly because this branch gets no CI (workflows only trigger on PRs into `main`), so that run is the only verification this PR will ever have.

Fixed in aa7a3c74.

PR #788, opened 2026-08-31
fix(workflows): a step says when its input is trying to steer it

ViaJables:
> **Stacked on #786** (base: `fix/extraction-prompt-injection`) — it introduces `injected_instructions`, which this reuses. Merge #786 first, or retarget this to `main` afterwards; only the top commit belongs to this PR.

## The ticket

A Prompt step asked *"What is the total award amount in this document?"* against an award letter showing **485,000 USD**. It answered **$1** — the figure a line planted in the document told it to report — and offered that as a completed deliverable, Download and Save to folder beside it. Nothing indicated the answer had been redirected.

## What was already true, and why it wasn't enough

Two defenses for this exist and neither is visible to the person reading the output:

- the grounded prompt every workflow LLM step uses refuses to treat its CONTEXT as instructions (#777), and
- invisible text is cut at the PDF reader before a step ever sees it (#777).

Both reduce how often the answer is wrong. Neither tells a user looking at "$1" why they should doubt it — and the visible-instruction variant (`qa-injection-test.pdf` prints its SYSTEM NOTE on the page) survives the scrub entirely.

## The fix

**Prompt, Formatter and Deep Analysis steps scan the text they are about to reason over** and finish with a warning quoting the passage when it contains text written at the model.

The scan sits on the **assembled context**, not on any one input, so it covers every source a step can read:

- the run's documents (the ticket's shape — a document selected as the workflow input),
- a document the step selects itself, and
- an earlier step's output — which is how a crawled page or an API response delivers the same payload into a step that never touched a PDF.

The warning uses the channel the run UI already renders, so the run reads **"Completed with 1 warning — check the output before using it"** with the passage underneath, instead of a clean green tick. No frontend change was needed.

**The step's answer is passed through unchanged.** Rewriting or blanking it would replace a visible wrong number with an invisible correction — the same failure in the other direction, and the thing the reporter is objecting to in the first place.

## Verification

8 new tests: the ticket's exact shape, a clean document staying clean, planted text arriving from a selected document and from an earlier step, Formatter and Deep Analysis, the empty-prompt failure still failing the way it did, and the step wrapper carrying the warning up to where the run banner reads it.

413 tests pass across the workflow and injection suites; `ruff` clean.

While writing these I had the first version of the Prompt test driving the node wrong — passing the document as the previous step's output while the step, with a `Document` trigger, reads `doc_texts`. It failed for the right reason and is worth knowing when reading the tests: which input source a step resolves to depends on the trigger.

## Also here

`describe_passages` grows an `advice` argument so the notice can end with what *this* surface did about it ("check this step's output against the document itself") rather than extraction's wording. The extraction caller keeps the default.


ViaJables:

ViaJables:

PR #790, opened 2026-08-31
fix(kb): strip instructions planted in a document out of what retrieval returns

ViaJables:
> **Stacked on #788** (base: `fix/workflow-injection-warning`, which is itself on #786). It reuses `injected_instructions` from #786. Merge order: #786 → #788 → this. Only the top commit belongs to this PR.

## The ticket

A document showing *Total Award Amount: 485,000 USD* was added to a knowledge base with text inside it stating the total is $1. Chat answered **$1**, with three citations pointing at that document.

The difference from the earlier reports in this family: **this one is not a one-time event.** The poisoned text is chunked, embedded, and served to every user who asks, on every ask, until someone removes the document. And the citations made the wrong answer look verified.

## The fix

**Retrieval strips passages written as instructions to the model out of every chunk it returns** — in `DocumentManager.query_kb`, `query_documents`, and the lexical `get_kb_chunks_containing`. That is the one seam chat, the workflow Knowledge Base step, KB validation and the verification preview all read through, so the planted text never reaches a model on any of them.

**Stripping at the read rather than at ingestion is the substance of this PR, not an implementation detail:**

- it protects knowledge bases that were poisoned *before* this shipped, with nothing to re-ingest and no migration;
- it does not depend on a model choosing to honour a system prompt that tells it to ignore instructions in documents. Chat has had that instruction for a while and it did not hold here — the reported answer came from `VandalAI-Fast`, and small fast models are exactly the ones that comply with a confident-sounding "SYSTEM NOTE" instead.

**In chat**, on top of the strip:

- a retrieved chunk that was *nothing but* planted text is dropped rather than cited — the citations are what made the false answer look verified, and a passage with no content left has no evidence to offer;
- the turn opens with a notice on the same channel that already reports a compacted context, because anything that changed what the model read has to reach the reader or the answer looks identical either way. It names the finding and says to check that answer against the document.

The notice panel in `ChatPanel` grows a heading for this case ("A document in this knowledge base tried to steer the answer:") so it doesn't render under "Context was compacted to fit the model:".

## Verification

8 new tests: the strip on a mixed chunk (real content kept, planted text gone), a chunk that is only planted text emptying out, a clean chunk untouched, `query_kb` returning sanitized chunks with the flag, a clean KB unchanged, and the three chat behaviours — planted-only chunk not cited, partly-planted chunk keeping its real content, clean KB producing no notice.

159 tests pass across the KB retrieval, chat, citation and pipeline suites; `ruff` clean; `tsc --noEmit` clean.

## What this does not do

Nothing tells the **KB owner** their knowledge base contains a poisoned document — the notice is per-answer, to whoever asked. A source-level flag on the KB's document list, set at ingestion, is the obvious follow-up and is where an admin would actually act on it.

Ingestion also still stores the planted text in the vector store; this removes it on the way out. Cleaning it at ingestion as well would shrink the blast radius for anything that reads Chroma directly, but would not help the KBs already poisoned, which is why the read side came first.


ViaJables:

ViaJables:

Issue #817, opened 2026-09-02
[Bug] Saving a model in the admin UI silently deletes its tokenizer and safety-margin overrides


arhyneRWU:
## Description

`PUT` on a model in the admin UI replaces the whole `available_models` entry with a fixed dict literal. Any key not named in that literal is destroyed. Three of the keys `context_budget` reads are not named in it, so **saving a model from the admin form silently deletes its tokenizer and margin configuration**, and the model reverts to `cl100k_base × 1.5` with no warning.

Found while scoping #816; it is independent of that work.

## Steps to Reproduce

1. Configure a model with an exact tokenizer — set `tokenizer_path` (or `tokenizer_cache_root`, or `token_safety_margin`) on its entry in `SystemConfig.available_models`. This has to be done directly in Mongo; see "Actual Behavior" for why.
2. Confirm it takes effect: `resolve_exact_tokenizer` returns a tokenizer and `token_safety_margin` returns `1.0` for that model.
3. In the admin UI, open that model, change anything at all — or nothing — and save.
4. Re-read the model's entry.

## Expected Behavior

The fields survive. Editing a model's display tier should not change how its tokens are counted.

## Actual Behavior

All three are gone. The model silently falls to rung 4 of the ladder — counted with `cl100k_base` and inflated by `DEFAULT_TOKEN_SAFETY_MARGIN = 1.5` — and nothing in the UI or the logs says a setting was dropped.

**Evidence.**

- `update_model` assigns a complete dict literal: `cfg.available_models[index] = { ... }` (`backend/app/routers/admin.py:1791-1812`). It is a replacement, not a merge, so any key absent from the literal is lost. The add path (`:1646-1668`) builds the same fixed set.
- `ModelAddRequest` (`:322-351`) declares 22 fields. `token_safety_margin`, `tokenizer_path` and `tokenizer_cache_root` are not among them.
- `grep -n "token_safety_margin\|tokenizer_path\|tokenizer_cache_root" backend/app/routers/admin.py` → **zero hits**.
- `grep -rn "token_safety_margin\|tokenizer_path\|tokenizer_cache_root" frontend/src/` → **zero hits**.

**What consumes the deleted keys.**

- `tokenizer_path` — `context_budget.py:238`, the explicit vocabulary for `resolve_exact_tokenizer`. This is rung 1 of the ladder, the exact-count path added in #648.
- `tokenizer_cache_root` — `context_budget.py:242`, the per-model HF cache override.
[...]

PR #818, opened 2026-09-02
feat(budget): measure the token margin against the provider instead of guessing 50% high

### LINKED ISSUE #816: [Feature] Provider-native token counting for hosted models — Claude and Gemini fall to the 1.5 guessed margin
> **Edited twice since filing.** The first edit fixed the framing. This one corrects the design after reading the code properly: **the margin does not become 1.0**, and the earlier claim that it would was wrong. Details in "What I got wrong" below. The goal is unchanged.

## Problem

This is **item 3 of the plan you set out in #629** on 17 Aug:

> 3. **For hosted models, use provider-native counting where it exists**, falling back to the current margin where it doesn't.

Items 1 and 2 shipped — exact local tokenization (#648, margin sized in #719) and the live self-check (#774). Item 3 is the piece still outstanding, restated on 21 Aug as *"the right follow-up for hosted models (its own issue when you're ready)."*

`token_safety_margin` (`backend/app/services/context_budget.py:285`) is a four-rung ladder:

1. a local `tokenizer.json` → `1.0`, exact (`resolve_exact_tokenizer:227`)
2. a configured `token_safety_margin` → that value
3. `_is_openai_model` → `1.0` — tiktoken *is* the vocabulary for `gpt-*` and the o-series (`:272`)
4. otherwise → `DEFAULT_TOKEN_SAFETY_MARGIN = 1.5` (`:57`), plus a warn-once

Rung 4 is the guess, sized in #719 for the ~45% worst case on digit-dense tables, so every model on it plans against a third less window than it has and trips #632 routing early. Anthropic and Google both publish an exact count, and both are on rung 4.

## What I got wrong in the first two versions

**The margin cannot become 1.0.** Three independent reasons, all in the existing code:

1. **The margin is applied per component, not per request.** `estimate_input_tokens:795-806` computes `Σ ceil(raw_i × m)` over the system prompt, user message, each history message, each document and attachment. `BudgetPlan.total_input_tokens:567` is a computed property summing those fields — there is no slot for a total that is not the sum of its parts.
2. **The planner recounts on mutated text.** The last-ditch trim loop `:981-1052` re-invokes the counter at `:994`, `:1011`, `:1047`, `:1050` on text it has just cut. A network counter cannot live in that loop, and a count of the original text is void once the text changes.
3. **A pre-plan count does not measure what is sent.** The wire payload is assembled at `chat_service.py:820-836`, *after* planning — segments wrapped in `--- BEGIN REFERENCE DOCUMENTS ---`, the system prompt travelling separately as `instructions`, history as `message_history`. That gap is what `REQUEST_SCAFFOLD_TOKENS` (`:94-115`) exists to absorb, and it has to stay.

**So the proposal is a measured margin, not an exact count.** Call the provider once for the payload, count the same text locally with tiktoken, and use `max(1.0, provider / local)` as that request's margin. Properties:

- It replaces a blanket 1.5 with something content-specific — nearer 1.02 on prose, 1.18 on a budget table, per this module's own measurement table at `:33-44`.
- **It survives compaction**, because a ratio stays approximately valid as content is trimmed, where a total does not.
- **It fails closed by construction.** No provider figure means no ratio, and the ladder returns 1.5 — not because a second code path remembered to, but because there is nothing else to return.
- Clamped at 1.0: a provider that counts *below* tiktoken cannot be used to reclaim window. This only ever reduces over-inflation.

## Proposed Solution

**Why the count stays out of `context_budget.py`.** Stronger than the "it's synchronous" argument I gave before: that module imports only `logging, math, dataclasses, functools, pathlib, typing, app/config.py, tokenizers, tiktoken` — no app service at all — while `llm_service.py` imports **it**. Reaching a provider client from inside it would be a circular import. It is a deliberate dependency leaf, and this design keeps it one.

So: count out of band in the already-async caller, and pass the provenance-carrying result in.

**The count itself needs no new plumbing.** `Model.count_tokens` is on the pydantic-ai base class; `AnthropicModel` and `GoogleModel` override it, `OpenAIModel` raises `NotImplementedError`, and `MeteredModel` (`llm_service.py:599`) is a `WrapperModel` that forwards it. So the object `get_agent_model()` already returns exposes it — no new credential path, no raw SDK, no new decryption call site.

**Routing has to be told the margin, not re-derive it.** `model_routing._sized_for:71-75` calls `token_safety_margin(current_name, current_config)` to recover the ruler its input was measured with. That function is pure over name and config and cannot know a native count arrived, so it would answer 1.5 for a figure measured at 1.05 and rescale by a factor never applied — understating the request at the routing boundary, in the hard-fail direction that function's own docstring exists to prevent. Fixing the planner without fixing `_sized_for` is a regression, not a partial win. The effective margin must travel with the number into `choose_document_model` / `suggest_document_model`.

## Alternatives Considered

- **Do nothing.** Safe, wasteful: a third of the window on Claude and Gemini, and earlier routing than necessary.
- **Substitute an exact total instead of a margin.** What the first version of this issue described. Rejected for the three reasons above — it cannot survive compaction, and it breaks `BudgetPlan`'s component sum and the `context_budget` SSE frame that reports it.
- **Make the planner async and count natively inside it.** Honest, and it would make every recount exact. Rejected: `plan_and_compact_context` counts once per segment inside a trimming loop, so this converts a bounded local computation into an unbounded number of network round trips per turn.
- **Set `token_safety_margin` per model by hand.** Works, but goes stale on the provider's release schedule — your objection in #629 — and per #817 the admin form deletes it on the next save.
- **Stored calibration (#629).** Declined, and I still think that was right.

## Additional Context

**The reach is narrower than I first implied.** `detect_api_protocol:372` routes any name containing `claude` to the *OpenAI* protocol unless `api_protocol` is set explicitly, and the OpenRouter / Insight-AI branches all produce `OpenAIModel`, which has no count endpoint. Gemini is the asymmetric case — `:377` auto-routes `gemini*` to `google`. So this fires for models an admin explicitly configured as `api_protocol: anthropic` or `google` with a direct provider key. The `ModelEditor` presets create exactly those, so it is a real population — but it is not every Claude deployment, and the PR should say so rather than imply a blanket fix. I have deliberately **not** proposed adding a symmetric auto-route for `claude-*`: that would change which client existing deployments use for every Claude model, which is a behaviour change well beyond token counting.

**Two traps worth naming before someone hits them.**

- **Gemini's count omits the system prompt.** `models/google.py:340` attaches `system_instruction` only when the provider is not `google-gla`, and `GoogleProvider(api_key=…)` is `google-gla`. Treating that figure as covering the request would silently drop several hundred tokens of system prompt — an under-count. Google needs a locally counted allowance added, or it does not qualify as exact.
- **The house HTTP pattern is the opposite of fail-fast.** `_get_loop_http_client` wraps `RateLimitRetryTransport`, which retries a 429 six times honouring `Retry-After` up to 60s, and `build_thinking_model_settings:520` sets a 120s timeout. Reused naively that is a two-minute pre-flight before the first token. The counter needs its own short timeout, closer to `model_probe.py`'s shape.

**What I can and cannot verify.** This deployment is entirely self-hosted, so every model resolves through rung 1 and nothing is on rung 4 — I have no Claude or Gemini model to exercise end to end. The parts that matter are testable without one: that a timeout, a 429, a `NotImplementedError` from an `OpenAIModel`, and a malformed body all yield no ratio and leave the margin at 1.5. Live verification against a real provider endpoint would have to come from a deployment that holds direct provider keys, and I will say that plainly in the PR's test plan rather than tick a box I cannot.

No urgency from me. I am building to the shape above; if you would rather it went differently, say so and I will rework it before it becomes a PR.

arhyneRWU:
## Summary

Item 3 of the plan you set out in #629: *"For hosted models, use provider-native counting where it exists, falling back to the current margin where it doesn't."* Items 1 and 2 landed as #648/#719 and #774.

`DEFAULT_TOKEN_SAFETY_MARGIN` is 1.5 because that covers the densest content anyone measured — a currency-dense budget table diverging ~45% from `cl100k_base`. Applied to everything, a hosted model answering prose plans against a third less window than it has, and crosses the #632 routing threshold early. Anthropic and Google publish exact counts, so where one is available the margin is now measured instead of assumed.

**It is a ratio, not a substituted total, and the margin does not become 1.0.** I said 1.0 in the first draft of #816 and that was wrong — three things in the existing code prevent it:

1. The margin is applied per component (`estimate_input_tokens:795-806`), and `BudgetPlan.total_input_tokens:567` is a computed property summing those parts. There is no slot for a total that isn't the sum of its components.
2. The planner recounts on mutated text inside its trim loop (`:981-1052`). A network counter can't live there, and a count of the original text is void once the text is cut.
3. A pre-plan count doesn't measure what's sent — the wire payload is assembled after planning, which is what `REQUEST_SCAFFOLD_TOKENS` exists to absorb.

So: count the assembled prompt, count the same payload locally, and use `max(1.0, provider / local)` as that request's margin. A ratio stays approximately valid as content is trimmed where a total does not.

## Changes

**`backend/app/services/native_token_count.py`** (new) — fetches a provider's count, or reports plainly that it couldn't. Gated on protocol before any I/O, since `OpenAIModel.count_tokens` raises and every OpenAI-compatible model would otherwise decrypt a key and build a client per turn to learn that. Five-second timeout, deliberately *not* the request's own settings — `build_thinking_model_settings` sets 120s and the shared client's `RateLimitRetryTransport` retries a 429 six times honouring `Retry-After` up to 60s, which would put a two-minute stall in front of every message. Blanket `except` by necessity: Anthropic raises `ModelHTTPError`, Google's `genai` errors are unwrapped, a misconfigured protocol reaches `OpenAIModel`; there's no complete and stable union of those. `CancelledError` passes through, being the caller leaving rather than a failure.

**`backend/app/services/context_budget.py`** — `NativeCount` plus a rung between the tiktoken-is-exact check and the 1.5 default, so it only ever displaces the guess. Rejected on model-name mismatch, zero baseline, non-positive count, or non-finite ratio, each falling through to the ladder. Clamped at 1.0: a provider counting below tiktoken may reduce over-inflation, never reclaim window. The "estimated, not exact" warning stays quiet only when a measurement actually supplied the margin. No new imports — the module stays a dependency leaf.

**`backend/app/services/model_routing.py`** — `_sized_for` re-derived the current model's margin to restate a request in a candidate's units. Handed a natively-counted number it would still answer 1.5 and divide by a factor never applied, understating the request by a third exactly where the router decides whether a candidate can hold it — the #648 defect at the routing boundary, which that function's docstring says it exists to prevent. Callers now state the margin their number was measured with.

**`backend/app/services/chat_service.py`** — the wiring. `_build_chat_prompt` is the existing assembly extracted verbatim, so the pre-flight counts the request chat is about to make rather than a reconstruction. The payload counted is the whole prompt including reference documents: in document chat the documents *are* the payload, and they're the digit-dense content the 1.5 was sized for — sampling question and history alone would measure prose and apply the answer to a budget table.

## Three things I'd want you to push back on

**The pre-flight is unconditional.** One serialized round trip ahead of the first token on every eligible turn. Gating it on "the estimate is near a boundary where the answer could change" would be cheaper, but it makes *is this count exact?* depend on request size, which is hard to reason about later. I have no latency data for either count endpoint — if you do, that's the number that should decide this.

**The reach is narrower than "Claude and Gemini".** `detect_api_protocol:372` routes a bare `claude-*` name to the OpenAI-compatible path, and the OpenRouter / Insight-AI branches produce `OpenAIModel`, which has no count endpoint. This fires only for models an admin explicitly configured `api_protocol: anthropic` or `google` with a direct provider key — which the `ModelEditor` presets do create. I deliberately did **not** add a symmetric auto-route for `claude-*`: that would change which client every existing Claude deployment uses, a behaviour change well beyond token counting.

**Google's count omits the system instruction.** pydantic-ai attaches `system_instruction` only for providers other than `google-gla`, and an api-key `GoogleProvider` is `google-gla`. Rather than fudge it, the system prompt is excluded from *both* sides of the ratio. Counting it on one side only would depress the ratio for a reason unrelated to tokenizer divergence — and a depressed ratio clamps to margin 1.0 over a tiktoken figure that itself under-counts these models, which is the hard-fail direction.

Also deliberately untouched: `stored_count_margin`. It answers "how safe is a count someone else took earlier, when I don't have the text", and a provider count requires the text — so `find_oversize_documents` and `find_context_overflow` keep the 1.5 allowance.

## Test Plan

- [x] Shared checks pass (`make ci`) — `make backend-ci` locally: 4438 passed, 161 skipped, coverage 64.38% against the 50% gate, tier-1 integration 14 passed. Baseline on this branch point was 4410 passed; +28 accounts exactly for the tests added. I could not run `make frontend-ci` locally (npm is not on my `PATH`) and said so when opening this; CI's **Frontend** job has since run and passed, along with the other ten checks. No `frontend/` file is touched by this PR.
- [ ] Release check (`make release-check`) — **not applicable.** No packaging or deployment path changed.
- [ ] Manually tested the affected feature(s) — **not possible here, and this is the PR's main limitation.** This deployment counts every model exactly from a local `tokenizer.json`, so nothing is on the guessed rung and there is no Claude or Gemini model to exercise. **No live provider count endpoint was called at any point.** What *is* covered, against stubbed providers: a good Anthropic count, a good Google count, `NotImplementedError` from an `OpenAIModel`, timeout, `ModelHTTPError`/429, a malformed usage object, and every ineligible protocol asserting that no client is constructed at all. Live verification needs a deployment holding direct provider keys.
- [ ] Updated deploy/release docs — **not applicable.** No install or release path changed.

## Related Issues

Closes #816.

Adjacent, filed separately and not addressed here: #817 — `update_model` replaces the whole model dict, so `token_safety_margin`, `tokenizer_path` and `tokenizer_cache_root` are deleted by the next save from the admin form. It constrains this work (any per-model opt-in would be wiped), which is why the pre-flight is gated on `api_protocol` — an existing field the form does manage — rather than on a new one.


ViaJables:
Reviewed and pushed one commit. This is careful work — the fail-closed design is genuinely thorough, the blast radius is narrow (`chat_service` is the only caller of `estimate_input_tokens`/`plan_and_compact_context`, so extraction and workflows take an identical path), and the routing change is right: `_sized_for` re-deriving 1.5 for a natively-counted number would understate the request by a third at exactly the boundary where the router decides whether a candidate can hold it. I verified the pydantic-ai signatures against 1.71.0, including the `google-gla` system-instruction omission.

**What I added: a circuit breaker.** The module coalesces the failure *log* but not the failure *work*. A deployment whose egress to the count endpoint is slow, or whose org is rate-limited there, pays the full 5s timeout on **every chat turn ahead of the first token, permanently** — and after the first occurrence says so only at DEBUG. No metric, no way out short of a restart.

Three consecutive failures now pause a model for five minutes. The check sits after the protocol gate (so an ineligible model never enters the circuit) and before any model construction, key decryption or request (so an open circuit costs nothing). One request goes through when the cooldown elapses; any success clears the record, so an outage cannot disable counting until the next deploy. Per model name, monotonic clock.

Skipping is strictly better than blocking here — the estimate path it falls back to is the one that shipped before this feature, so a paused model gets the same budget it would have got anyway, without the wait.

**Two changelog corrections**, both things the entry asserted that the code does not do:

- *"fires only for models an admin explicitly configured"* is not true of Gemini. `detect_api_protocol` routes any name containing `gemini` to `google` on its own (`llm_service.py:377`), so a deployment already running one gets the pre-flight on upgrade without changing anything. Issue #816 stated this correctly — the entry lost it.
- *"one unconditional round trip"* is no longer true, which is the point of the commit.

Your existing failure-logging tests needed updating rather than working around: `_CIRCUIT` is per-process state exactly like `_UNAVAILABLE_LOGGED`, so it joins that autouse reset — otherwise whether a test sees an open circuit depends on collection order. I also extended the repeated-failure test from 4 turns to 50, so it pins the bound rather than an arithmetic coincidence.

## Two things I did not change

**The payload is uploaded twice per turn.** `_build_chat_prompt` is passed the pre-compaction segments, so a 150k-token document chat POSTs ~600KB to `count_tokens` and then the same ~600KB to `/v1/messages`. That is correct for measurement — the documents are the digit-dense content the 1.5 was sized for — but it doubles egress and pre-request latency on exactly the largest requests, which is also where the 5s budget is most likely to blow. Worth considering whether a sampled or capped payload would measure well enough.

**A merge-forward hazard for `major/agentic-chat`.** `_ask_provider` passes a bare `ModelRequestParameters()`, justified by "the chat agent registers none" — true on `main`, false on the v5 branch, where `create_agentic_chat_agent` registers the `chat_tools` registry. Merged forward, the provider would count a payload missing thousands of tokens of billable tool schemas, the ratio deflates toward 1.0, and the 1.5 cushion incidentally covering those schemas disappears — in the under-estimate direction. Nothing in the code or tests would flag it at merge time. Worth a guard or at least a comment before that merge.

ViaJables:
Merged after review. Two things the review found that were not blocking, recorded here as follow-ups:

1. **The pre-flight counts the whole uncompacted payload on every eligible turn** (`chat_service.py`, the `_measure_native_count` call around the prompt build). That is a second full upload of every document segment ahead of the first token, and for a payload already above the model's window the count endpoint may reject it — three such turns open the circuit for five minutes for every user of that model, which is exactly the large-document case the margin matters for. Worth skipping the pre-flight when the local estimate already fits with headroom or exceeds the window outright, or capping the counted payload and applying the ratio.
2. `native_token_count._ask_provider` builds a fresh provider client (and decrypts the key) per count; passing the model instance the chat is about to build would halve that.

Committed on the branch before merge: the boolean-margin warning in `model_routing._measured_margin` now logs the offending value instead of `None`, and the CHANGELOG entry was moved back under `[Unreleased]` (git had auto-merged it into the released v4.12.0 section).

Issue #819, opened 2026-09-02
[Feature] Separate the tool's name from the institution's name — one field currently does both


arhyneRWU:
## Problem

`org_name` (`backend/app/models/system_config.py:212`) is doing two different jobs, and a white-labelled deployment cannot satisfy both at once.

**As the tool's name**, it is the word users read in conversation:

- `ChatPanel.tsx:174` — `Hi, I'm your ${branding.orgName} assistant.`
- `ChatInput.tsx:162` — `Ask ${branding.orgName} anything...`
- `ChatPanel.tsx:581` — `Welcome to {branding.orgName}`
- `router.tsx:95` — the browser tab title
- `Landing.tsx:379` — the screen-reader `h1`

**As the institution's name**, it is a data field that outlives the session:

- `verification_service.py:313-323` — the static creator credit, whose own comment reads *"by Jane Doe at University of Idaho ... so attribution survives catalog export/seeding to installs where the submitter's user account doesn't exist."*
- `export_import_service.py:658` — `deployment_org` on exported items
- `demo_service.py:317` — team naming, `f"{org_name} - {department}"`
- `email_service.py:122` — substituted into outgoing mail
- `Footer.tsx:13` — the footer heading

Set it to the institution and the chat says *"Hi, I'm your Roger Williams University assistant"* and *"Ask Roger Williams University anything..."*, which reads as though the university is the chatbot.

Set it to the tool's name and it is worse than awkward: **the creator credit stamped onto every catalog item this deployment contributes now names a product instead of an institution** — and that credit is exported and seeded into other people's installs, where nobody can tell it was a branding setting rather than a real affiliation. The one field that is supposed to make attribution survive export is the same field a deployment must overload to get a sensible chat greeting.

There is no setting that is correct for both.

## Proposed Solution

Add a second, optional field. Nothing renamed, nothing removed.

```python
# system_config.py
org_name: str = ""   # unchanged — the institution
app_name: str = ""   # new — what the tool calls itself; falls back to org_name
```

Resolution order is `app_name or org_name or "Vandalizer"`, so:

- **A deployment that sets neither** behaves exactly as today.
- **A deployment that has already set `org_name`** behaves exactly as today — `app_name` is empty, so every surface still reads `org_name`. This is the important property: no existing install changes appearance or data on upgrade.
- **A deployment that sets both** gets the institution on institutional surfaces and the tool's name in conversation.

[...]

PR #821, opened 2026-09-02
fix(branding): a white-labelled sign-in page shouldn't show our gold button

arhyneRWU:
## Summary

A white-labelled deployment currently shows **its own logo above a Vandalizer-gold button** on the sign-in page — the first screen any new user sees. Same on password reset and both invite-acceptance flows. Those surfaces hardcode `#f1b300` instead of reading the configured brand colour.

`pages/Landing.tsx` was migrated for exactly this reason in `1b24af2f` ("the landing page hardcoded Vandalizer's gold … so unauthenticated visitors always saw the default palette regardless of what admins configured"). These surfaces were missed. This follows that precedent rather than proposing a new idea.

## The prerequisite, and why it's in the same PR

Migrating naively would have made things **worse**. `getAccessibleOnLight` darkens the brand colour until it passes contrast on white; nothing did the equivalent for the near-black auth and footer surfaces (`AuthLayout.tsx:8`, `Footer.tsx:7` are `bg-[#0a0a0a]`).

Measured on `#0a0a0a`:

| | ratio |
|---|---|
| `#f1b300` — today's hardcoded gold | 10.55:1 |
| `#163A64` — a navy brand, used raw | **1.72:1** |
| `#367ed3` — same navy via `getAccessibleOnDark` | 4.79:1 |

So `getAccessibleOnDark` lands first (lightens rather than darkens, holds hue — 212.3° → 212.5°), exposed as `--color-highlight-on-dark`. The two halves are in one PR because the second is incorrect without the first; the first commit touches no component and is inert on its own.

Related: `ThemePanel` offers no contrast warning at all, so an admin can pick `#1a1a2e` today and nothing objects. This doesn't fix that, but it makes the choice survivable.

## Changes

Seven files, colour classes only — no copy, layout or structure: `LoginForm.tsx` 8, `ResetPassword.tsx` 16, `InviteAccept.tsx` 22, `JoinLinkAccept.tsx` 24, `JoinProjectAccept.tsx` 4, `AuthLayout.tsx` 1, `Footer.tsx` 5.

Every `text-[#f1b300]` was classified by walking up to its nearest painted background rather than inferred from the filename. All 14 resolved to on-dark; the two inside `bg-[#f1b300]/10` tint discs were checked against the card beneath (`bg-[#171717]`) and the page under that.

**Seventeen buttons paired `bg-[#f1b300]` with hardcoded `text-black`.** Those move to `text-highlight-text` — the runtime-computed contrast colour — in the same edit. Changing the background alone is how black text ends up on navy.

## Deliberately not changed

- **The footer's attribution.** "Powered by Vandalizer", the GPL link and the NSF GRANTED acknowledgement are byte-identical; the diff for that file is five `className` strings. The obligation is the words, not the hover colour.
- **Two SSO buttons** keep `text-black` — they sit on `bg-white`, not on brand.
- **One `hover:text-[#d49e00]`** (`ResetPassword.tsx:38`) stays hardcoded. The obvious substitution is `--highlight-hover`, but that is a *darkened* brand colour and would read worse on near-black than what it replaces. A correct fix needs an on-dark hover variant that doesn't exist; I'd rather leave one honest hex than introduce a wrong token.

## Behaviour change worth stating

On an unbranded deployment these surfaces shift from `#f1b300` to `#eab308` once the theme resolves — the CSS fallback and the backend default (`system_config.py:204`) were always two slightly different golds. Visually near-identical, no contrast consequence, but it is a change rather than a no-op, and `--highlight-on-light` has the same property today.

## Test Plan

- [x] Shared checks pass (`make ci`) — frontend **746 passed / 102 files** (this branch's base is 737 on `main`; +9 is 8 colour-utility tests plus one `LoginForm` assertion). `tsc -b` exit 0. `eslint .` 0 errors, 20 pre-existing warnings untouched. Backend unaffected — no backend file changed. There is no `node`/`npm` on my machine, so the frontend legs ran in a pinned, network-isolated container against the committed `node_modules`; no `npm install`, no lockfile change.
- [ ] Release check (`make release-check`) — **not applicable.** No packaging or deployment path changed.
- [ ] Manually tested the affected feature(s) — **not run, and this is the PR's weak point.** I have no browser here, so the on-dark classifications are reasoned from markup, not observed. Worth noting the suite is close to blind to colour: there are no snapshot tests, no `toHaveStyle`, and no colour assertions in the Playwright specs, so **these tests would pass even if the migration were wrong**. I added one meaningful assertion rather than manufacturing coverage that would imply verification I don't have. A reviewer loading the sign-in page on a branded deployment is the real check.
- [x] Updated `CHANGELOG.md`
- [ ] Updated deploy/release docs — **not applicable.**

## Related

Part of a larger pattern: ~244 static brand-colour literals remain across the app, plus three drifting neutral grey scales with several pre-existing WCAG AA failures. Happy to open an issue with the measurements if useful — this PR is the highest-value slice, not the whole problem.


ViaJables:
Reviewed and pushed one commit to this branch (thanks for `maintainerCanModify`).

**The fix:** these auth inputs set `focus:outline-none`, which suppresses the global `:focus-visible` outline, so the ring *is* the focus indicator. Painted with the raw brand token, a dark brand (`#163A64` over `#0a0a0a`) computes to roughly **1.3:1** — a keyboard user cannot see where they are. Gold was visible, so the migration regressed focus visibility on exactly the deployments it exists to serve. 14 focus rings and 3 spinner borders now use `--highlight-on-dark`.

Also: `AuthLayout` still had `bg-yellow-600/10` — the blurred wash across 40% of the viewport behind every auth page. The audit grepped for `#f1b300`, and named Tailwind yellows did not match, so a navy sign-in page kept a large yellow halo.

The `getAccessibleOnDark` maths is correct — I recomputed `contrastRatio(#eab308, #0a0a0a) = 10.3`, so the default gold really does pass through unchanged.

**One thing worth your call before this merges.** `applyTheme` sets `--highlight-on-dark` unconditionally, so every existing deployment shifts from `#f1b300` to `#eab308` on the auth pages. Harmless (the two golds are nearly identical, no contrast consequence) and you disclosed it — but #824 introduces `THEME_DEFAULT_COLOR` precisely to keep the default path untouched, and this PR does not use it. The two halves of the same stack guard the default differently. Three lines here would make unbranded installs see literally no change.

**Left alone deliberately:** `ProtectedRoute.tsx:14` has the same raw-brand spinner, but its container declares no background, so I could not confirm it renders on dark. Flagging rather than changing on a guess.

On test strength — the suite had one assertion across fourteen migrated components, which is why this was structurally undetectable. The case I added asserts the on-dark token is present **and** the raw one absent, since the raw token is right on light surfaces and wrong only here. Verified by reverting the component: both cases go red, green again with the fix.

ViaJables:
Merged after review. One follow-up, not blocking:

- The primary buttons on the `#0a0a0a` auth surfaces (`LoginForm`, `ResetPassword`, both invite pages) still use raw `bg-highlight`. The *text* on the button is handled by `text-highlight-text`, but the button's silhouette is not: a dark brand such as navy `#163A64` gives about 1.7:1 button-vs-page, under the 3:1 non-text minimum. Either paint the button with `bg-highlight-on-dark` and add a `--highlight-on-dark-text` token computed from it, or give the button a visible border.

Committed on the branch before merge: the union of this branch's `BrandingContext.test.tsx` block with #819's block from main (both appended a `describe` at the end of the file).

PR #824, opened 2026-09-02
feat(branding): the dark app chrome follows the brand colour

arhyneRWU:
> **Stacked on #821.** This branch contains that PR's commit as well; review only `e472d211`, or merge #821 first and this becomes a single-commit diff. Both touch `color.ts`, `BrandingContext.tsx` and `index.css`, which is why they aren't independent.

## Summary

`--color-panel-dark` is a fixed `#191919` in `@theme` that nothing derives from anything. So a deployment branded to navy gets navy accents sitting on neutral-black chrome — the workspace tab strip (`RightPanel.tsx:31`), the document-panel header (`LeftPanel.tsx:242`, commented "Black header bar"), the knowledge panels, the left rail. The brand stops at the accent colour and the shell stays default.

It's now derived from the brand's hue and saturation **with the lightness pinned**.

## Why pinning lightness is the whole design

Tinting neutrals toward a brand colour is normally a bad idea: it makes contrast a runtime property of a value you don't control, and `ThemePanel` offers no contrast warning at all — an admin can pick `#1a1a2e` today and nothing objects.

Pinning the lightness sidesteps that. Only hue and saturation vary, so the chrome/text contrast is bounded by construction:

| brand | derived chrome | vs white text |
|---|---|---|
| — (current) | `#191919` | 17.58:1 |
| `#163A64` navy | `#0f1d2e` | 16.99:1 |
| `#eab308` gold | `#2e260f` | 15.0:1 |
| `#808080` grey | `#1f1f1f` | neutral, saturation 0.0000 |
| `#ff0000` red | `#2e0f0f` | saturation capped at 0.508 |

Saturation is capped so a fully saturated brand doesn't read muddy; a greyscale brand comes back genuinely neutral rather than faintly tinted.

## Unbranded deployments are unchanged

When `highlight_color` is still the shipped default, the variable is **removed** rather than set, so the `:root` neutral wins and the chrome is pixel-identical.

`removeProperty` rather than an early skip, deliberately: `applyTheme` runs once for the cached theme and again for the server's, so an admin resetting a brand back to the default has to clear the value this same call previously wrote. There's a test for that path, and it's what keeps the guard tests from being vacuous — it asserts navy is set first, then cleared.

This mirrors `_THEME_DEFAULT_COLOR` in `backend/app/services/email_service.py:104-107` ("if highlight_color still equals this, the admin hasn't picked a brand color, so leave the email gold alone"), cited in the comment so it reads as an existing convention rather than a special case.

## Changes

`getPanelDark()` in `color.ts`, one `setProperty`/`removeProperty` branch in `BrandingContext`, the `@theme` entry moved to the `var(runtime, fallback)` form, and **13 hardcoded `#191919` backgrounds migrated onto the token** so the chrome can't end up half-branded.

Seven live literals are deliberately left:
- **Both certification-badge gradients** (`CertifiedBanner.tsx:9`, `ActivityRail.tsx:439`) — `linear-gradient(135deg, #191919, #2d2d2d)`. The second stop is out of scope, so tinting only the first gives a navy→neutral-grey hue break, and migrating one badge but not its twin would desynchronise them. Doing these properly needs a second derived stop.
- **Four text colours** that merely share the hex (`FileList.tsx:110/129`, `FolderRow.tsx:161`, `FileRow.tsx:218`) — tokenising a glyph colour to a chrome *background* variable would be wrong.
- **One 8-digit alpha form** (`ChatInput.tsx:149`, `#19191913`) needing `color-mix()` rather than a swap.

One judgement call worth a reviewer's eye: `ChatMessage.tsx:287` is a content surface rather than shell chrome. It's included because that bubble already carries `borderLeft: 7px solid var(--highlight-color)` — it's brand-aware already, so tinting its fill is coherent. Say the word and it comes out.

## Test Plan

- [x] Shared checks pass (`make ci`) — frontend **756 passed / 102 files** (746 at branch start; +10 is 6 colour-utility tests and 4 `BrandingContext` tests). `tsc -b` exit 0. `eslint .` 0 errors, same 20 pre-existing `react-hooks/exhaustive-deps` warnings, count unchanged. No backend file touched. No `node`/`npm` on my machine, so the frontend legs ran in a pinned, network-isolated container against the committed `node_modules` — no `npm install`, no lockfile change.
- [ ] Release check — **not applicable.**
- [ ] Manually tested — **not run.** No browser here. The contrast figures are computed and asserted, but whether navy chrome *looks better* than neutral black is a taste question a screenshot answers and I can't. If you think branded chrome is the wrong call for this product, that's a reasonable position and this PR is the wrong idea rather than a wrong implementation.
- [x] Updated `CHANGELOG.md`
- [ ] Updated deploy/release docs — **not applicable.**


ViaJables:
Merged after review. Two things committed on the branch before merge, and one follow-up:

- **Committed:** `applyTheme` now treats anything that is not a `#rrggbb` colour like the default and clears `--panel-dark`. `getPanelDark('')` produced a non-colour, and `var(--panel-dark, #191919)` does not fall back for an *invalid* value, so a malformed stored theme painted the whole shell transparent. Test added.
- **Committed:** CHANGELOG entry moved from `### Fixed` to `### Changed` — the chrome following the brand is new behaviour, not a bug fix. Also re-applied the `WorkflowEditorPanel` swap at main's new location of the "Add task" button.
- **Follow-up:** the on-dark accent from #821 is computed against `#0a0a0a`, not the tinted chrome, so with a navy brand, navy accents on navy-tinted chrome get slightly *less* distinguishable than on `#191919`. A follow-up should compute an on-chrome accent.

PR #827, opened 2026-09-02
fix(optimizer): orphaned workflow/extraction runs are reaped instead of blocking forever

### LINKED ISSUE #814: Extraction and workflow optimizer runs have no orphan janitor; a hard-limit kill blocks re-optimization permanently
**Evidence.** `kb_optimization_janitor` exists for KB runs only (`celery_app.py:124-127`, `kb_validation_tasks.py:270-334`). `optimize_extraction_task` (`extraction_tasks.py:334-340`) and `optimize_workflow_task` (`workflow_optimization_tasks.py:24-32`) carry the same `soft_time_limit=5400 / time_limit=5460` and have **no equivalent reaper**.

**Soft limit is handled** — `extraction_optimizer.py:798-806` and `workflow_optimizer.py:494-510` both catch `Exception` and set `status="failed"`. A `SoftTimeLimitExceeded` is recorded correctly.

**The hard limit is not.** At 5460s the child is SIGKILLed, no handler runs, and the run document is left at `status="running"` forever. `routers/extractions.py:2177` then returns `409` on `active_run`, so the search set is **permanently blocked from re-optimizing** until someone edits the database by hand. The KB janitor's own docstring (`kb_validation_tasks.py:273-277`) describes exactly this consequence as the reason it was written.

**Fix.** Generalize `kb_optimization_janitor` over all three run types, or add the equivalent sweep for `ExtractionOptimizationRun` and `WorkflowOptimizationRun`.

ViaJables:
Closes #814. Companion to #825 (which reaps stuck *workflow runs*; this reaps stuck *optimizer runs*).

## What was broken

All three optimizers handle the **soft** time limit honestly (broad `except Exception` marks the run failed). The **hard** limit (5460s SIGKILL) or a dead worker runs no handler at all, and coverage was uneven:

- KB runs: `kb_optimization_janitor` existed.
- Extraction runs: healed by `reap_one` via reap-on-start and the read endpoints.
- `WorkflowOptimizationRun`: **nothing** — and `start_workflow_optimization` 409s on any non-terminal run with no preceding sweep, so a hard-killed run didn't just spin: it **permanently blocked re-optimizing that workflow** until someone edited the database. The KB janitor's own docstring names this exact consequence as its reason for existing.

## The fix

- **The janitor is generalized over all three run collections** and renamed `tasks.passive.optimization_janitor` (beat key `optimization-janitor`, hourly as before). KB and workflow runs share the finalize-as-failed shape — no Celery task id on those docs to revoke, so finalizing the document is the whole job; workflow runs also get `stopped_reason="failed"`, matching every other terminal write on that model. Extraction candidates are **delegated to `extraction_optimizer.reap_one`** rather than duplicated: it also revokes the Celery task and honors the cancel grace window, and it no-ops on runs young enough to still be legitimately executing.
- **`workflow_optimizer` gains `reap_one` / `reap_stale_runs`** (cutoff = hard time limit + 4 min slack, mirroring extraction's `STALE_RUN_TIMEOUT_SECONDS`), and **`start_workflow_optimization` sweeps before its active-run check** — the same shape as `start_extraction_optimization` — so the block clears on the next click rather than the next hour.

## Testing

12 tests in the renamed `test_optimization_janitor.py`: the original KB sweep assertions, workflow runs reaped with `stopped_reason`, extraction delegation (both the no-op and finalized cases counted correctly), cutoff/status query shape, `reap_one` stale/young/terminal behavior, and start-path sweep-before-409. 11 of 12 verified red against the unfixed code. Full backend unit suite: 4653 passed, 96 skipped. Ruff clean.


ViaJables:
Retargeted to main (with #825, #826, #828–#830 — all six bugs exist on main, not just the v5 branch). One adaptation: main has no `optimization_actions.py` — the start path lives in `routers/workflows.py` — so the reap-on-start sweep is wired into the router's optimize endpoint there, and the start-path test is now an endpoint test (409 still fires for a genuinely active run, but the sweep is verified to have run first). The v5 branch keeps its `optimization_actions.py` version and will reconcile on its next catch-the-base merge.

PR #829, opened 2026-09-02
fix(citations): no pages on partial OCR; legacy interpolated markers hedge again

### LINKED ISSUE #807: Interpolated page *positions* are wrong on partially-converted OCR text; legacy markers read as exact
SCOPE NOTE: this is about marker **positions**, not quote matching. `extraction_sources.py` was audited after this issue was filed and the matcher is sound — do not rewrite it. `find_quote_offset` (`:145-166`) does exact-then-normalized matching with index-map projection; `normalize_with_map` (`:114-142`) folds smart quotes, dashes, NBSP and ligatures; `page_marker_for_offset` (`:169-182`) deliberately returns the marker so `approximate` survives (propagated at `:520-521`); `support_state` (`:60-91`) separates "quote exists" from "quote contains the value", with `_match_boundaries_ok` (`:285-307`) stopping 5000000 from supporting 500000.

The defect is upstream of all that: a correctly-located quote still resolves to an interpolated page.

**Evidence.** `_interpolate_page_markers` (`document_readers.py:687-714`) spreads pages uniformly by character count (`step = len(text) // num_pages`, `:705-708`) — a budget page and a references page differ by an order of magnitude. Worse, it interpolates against `pdf_page_count` (`:886`), the *source PDF's* count, while the text may cover a fraction of it (partial conversion, `ocr_client.py:270-314`). Spreading 400 page numbers over text covering 30 is systematically wrong, and `p. ~N` (`page_locator.py:44-52`) does not say so. Additionally `page_locator.py:13-14`: markers persisted before the `approximate` flag **read as exact**, so pre-existing documents render confident `p. 234`.

**Fix.** Suppress page citations (fall back to document title) when `partial_ocr` is set; backfill or invalidate pre-flag markers.

ViaJables:
Closes #807.

Per the issue's scope note (and the follow-up comment repeating it): the quote→page **matcher** in `extraction_sources` is sound and is untouched here. This PR fixes the marker **positions** underneath it.

## Fix 1 — partial OCR gets no page markers

`_interpolate_page_markers` spreads the **source PDF's** page count evenly across whatever text OCR returned. When docling reports `partial_success`, the text covers an unknown fraction of those pages — 400 page numbers spread over text from 30 pages is systematically wrong on every marker, and `p. ~N` doesn't say "possibly 370 pages off". `_read_pdf_text_and_markers` now suppresses page markers entirely when the OCR report says partial (main path and the PyMuPDF-failed fallback that re-ships partial text), so citations fall back to the document title. No page beats a wrong page. The reader keeps a local report dict when the caller passes none, so the signal exists regardless.

## Fix 2 — legacy pre-flag interpolated markers hedge again

Markers persisted before the `approximate` flag existed carry no key and read as exact — the NOAA proposal's 79 interpolated markers rendering confident `p. 234`. The interpolator has always placed page N at exactly `N * (len(text) // num_pages)`, so **perfectly uniform spacing across ≥ 3 page markers is its signature**; measured boundaries essentially never produce three pages of identical character length, and a rare false positive costs only an unnecessary `~` (that asymmetry is what the heuristic leans on; 2-page documents are left alone — one delta proves nothing).

`page_locator.with_marker_provenance` applies this wherever stored markers are consumed: chat (`_page_positions`, `_has_approximate_pages` — covering annotation, page notes, and derived citation chips), extraction sources, and chunk metadata in `DocumentManager` (both document and KB ingestion), so a re-ingest/refresh backfills hedged chunk metadata for legacy documents. This is backfill-on-read — no migration of stored documents or ChromaDB required.

## Fixture honesty note

Several existing test fixtures modeled "measured" boundaries with uniform offsets (`n * 100`) — which is precisely the interpolator's signature, so the detector (correctly) flagged them. Those fixtures now use irregular offsets, and `test_page_notes`'s `test_markers_written_before_the_flag_existed_are_treated_as_measured` — which documented the NOAA false-negative as accepted behavior — is replaced by a pair pinning the new split: uniform pre-flag markers hedge, irregular pre-flag markers still read as measured.

## Testing

15 new/updated tests across `test_page_locator` (detector behavior: legacy flagged, measured untouched, already-flagged/2-page/sheet/None pass-through, mixed lists), `test_document_readers_helpers` (partial → no markers with report preserved, partial with no caller report, full conversion still interpolates hedged markers), and `test_page_notes` (the flipped legacy pair). 9 verified red against the unfixed code. Full backend unit suite: 4656 passed, 96 skipped. Ruff clean.


ViaJables:
Retargeted to main. One adaptation: main's `annotate_pages` inlines the positions loop (`_page_positions` and `derive_document_citations` are v5-branch-only), so the legacy-marker normalization is applied at the top of `annotate_pages` instead. Everything else transplanted cleanly. The v5 branch reconciles on its next catch-the-base merge.

PR #842, opened 2026-09-03
feat(workflow): default model for all steps (canvas selector)

### LINKED ISSUE #841: Workflows: no way to set one model for all steps
### Problem

When a workflow runs without an explicit model, every step falls back to the caller's global default model (`get_user_model_name`). The only way to make a whole workflow run on a specific model is to set a per-step **Model Override** on each step individually.

For multi-step workflows (e.g. an *RFA checklist*, 6+ tasks) this is tedious, easy to get inconsistent, and there is no single place to see or change the model a workflow runs on.

### Proposal

Add a **workflow-level default model**, set once on the Design canvas, applied to every step that doesn't have its own override.

- A **"Model — all steps"** selector on the Design canvas.
- Persisted on the workflow (`input_config.default_model`).
- `run_workflow` / `run_workflow_batch` use it when no explicit model is passed.
- **Precedence:** explicit run model > workflow default > user / system default.
- A per-step **Model Override** still wins.
- **"Automatic (system default)"** (blank) keeps today's behaviour.

### Acceptance criteria

- [ ] Selector visible on the Design canvas; disabled when the user can't manage the workflow.
- [ ] Selecting a model persists to `input_config.default_model` and survives a reload.
- [ ] A run with no explicit model uses the workflow default; blank falls through to the system default.
- [ ] Per-step overrides unaffected; existing workflows unaffected (backward compatible).

StringTheoryDev:
Closes #841

## What

Adds a workflow-level default model — a **"Model — all steps"** selector on the Design canvas — applied to every step that has no **Model Override** of its own.

## Why

Running a multi-step workflow on a specific model previously required setting a per-step override on every step. This adds one place to pick the model the whole workflow runs on.

## How

- **Backend** (`workflow_service.py`): in `run_workflow` and `run_workflow_batch`, when no explicit model is passed, fall back to `wf.input_config.default_model` before the user default. The resolved model is stamped on the `WorkflowResult` and passed to the step engine, which already treats it as the per-step default (a step's own override still wins).
- **Frontend** (`WorkflowEditorPanel.tsx`, `workflow.ts`): a canvas selector that persists to `input_config.default_model` (optimistic save), populated from `getModels()`. "Automatic (system default)" = blank = prior behaviour.

## Precedence

explicit run model → workflow `default_model` → user / system default. Blank is never used as a model.

## Testing

- `backend/tests/test_workflow_default_model.py` — 5 unit tests covering the full precedence (explicit wins; workflow default used and short-circuits; missing / null / blank fall through).
- Verified live end-to-end against a running stack: set on the canvas → persisted to Mongo (`input_config.default_model`) → resolved by `run_workflow` → workflow restored to its prior state.
- `ruff` (0.15.7) clean on `app/`; `tsc -b` clean; workflow test suite (82 tests) green.

## Compatibility

Additive and backward compatible — existing workflows keep the system default until a model is chosen.

### DIFF
--- a/backend/app/routers/workflows.py
+++ b/backend/app/routers/workflows.py
@@ -1114,8 +1114,17 @@ async def test_step(request: Request, req: TestStepRequest, user: User = Depends
             ss = await get_authorized_search_set(search_set_uuid, user)
             if not ss:
                 raise HTTPException(status_code=404, detail="Search set not found")
+    # Authorize the workflow before reading anything off it, so this cannot
+    # be used to learn another tenant's configured model.
+    workflow_input_config = None
+    if req.workflow_id:
+        wf = await get_authorized_workflow(req.workflow_id, user)
+        if not wf:
+            raise HTTPException(status_code=404, detail="Workflow not found")
+        workflow_input_config = wf.input_config
     task_id = await svc.test_step(
-        req.task_name, req.task_data, document_uuids, user.user_id, req.model
+        req.task_name, req.task_data, document_uuids, user.user_id, req.model,
+        workflow_input_config=workflow_input_config,
     )
     return {"task_id": task_id}
 
--- a/backend/app/schemas/workflows.py
+++ b/backend/app/schemas/workflows.py
@@ -135,6 +135,11 @@ class TestStepRequest(BaseModel):
     task_data: dict
     document_uuids: list[str]
     model: Optional[str] = None
+    # Lets Test Step honour the workflow's own default model. A step whose
+    # selector reads "Use workflow default" was otherwise tested on the user's
+    # default and run on the workflow's -- tuned against one model, run on
+    # another, with the label saying otherwise.
+    workflow_id: Optional[str] = None
 
 
 class ReorderStepsRequest(BaseModel):
--- a/backend/app/services/workflow_optimizer.py
+++ b/backend/app/services/workflow_optimizer.py
@@ -1169,8 +1169,14 @@ async def _execute_workflow_inproc(
     sys_config = await SystemConfig.get_config()
     sys_config_doc = sys_config.model_dump() if sys_config else {}
 
-    # Default model: same resolution path Celery uses.
-    model = await get_user_model_name(user_id)
+    # Default model: same resolution path Celery uses -- the workflow's own
+    # default first, then the user's. Measuring the workflow on a model it will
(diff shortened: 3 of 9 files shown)


ViaJables:
Reviewed and pushed one commit. Thanks for this — the core mechanism is right: precedence is correct, a per-step Model Override still wins, and existing workflows with per-step models are genuinely untouched on both save and run. I checked that specifically.

What I changed: the canvas says a default model **"Runs every step on this model"** and does not qualify it, but only the interactive and batch run paths honoured it. Three others did not:

- **Scheduled / folder-watch runs** (`passive_tasks.py`) resolved `available_models[0]` and never looked at the workflow — it did not even honour the owner's default. Set the default to Sonnet, click Run → Sonnet; let the nightly automation fire → whatever is first in System Config. Same workflow, two models, nothing in the UI saying so.
- **The optimizer's in-process run** resolved the user default under a comment reading *"same resolution path Celery uses"* — a comment this feature made false. Validation was scoring the workflow on a model it would not run on, so the score, and any prompts or overrides tuned from it, described a configuration that never executes.
- **Test Step** took the user default, so a step whose selector reads "Use workflow default" was tuned against one model and run against another. That needed the workflow threaded through the schema, router and canvas call; the authorization check runs before `input_config` is read.

The resolution was already duplicated between the interactive and batch paths, so it now lives once in `resolve_run_model()`. Precedence is unchanged.

On tests: the five new cases all fail with the app changes reverted. One covers the **batch** path specifically — its copy of the resolution line was independently revertible with zero coverage, and batch is the default whenever more than one document is selected, which is the multi-document case #841 is really about.

**One thing I deliberately did not touch, because it is a product call rather than a bug.** The agentic-chat `run_workflow` tool passes the ambient chat model as `model=`, which makes it look like an explicit request and outrank the workflow default. So "run the RFA checklist on this document" in chat uses your chat model while the canvas still reads the workflow's. Whether that is wrong depends on whether the chat's model should win — worth deciding before this merges forward to `major/agentic-chat`, where chat is the common run path.

**Also pre-existing, not from this PR:** if an admin later removes the chosen model from System Config, the canvas falls back to displaying "Automatic (system default)" while runs still resolve the dead name and fail at the provider. `input_config.default_model` bypasses `get_user_model_name`'s tag resolution and self-heal. Worth a follow-up.

PR #853, opened 2026-09-03
fix(ingestion): four fidelity papercuts between what a file shows and what we read

### LINKED ISSUE #812: Ingestion fidelity papercuts
- `clean_markdown_nans` does a blind `cleaned.replace("NaN", "")` (`document_readers.py:45`) — corrupts legitimate strings containing "NaN" (a PI surname, "NaNoparticle").
- Plain-text branch uses bare `encoding="utf-8"` with no fallback (`:965-967`) while `_read_text_with_fallback` (`:575-584`) exists in the same file and is used only by the CSV viewer — so a cp1252 CSV *fails* while an unidentifiable binary *succeeds*.
- XLSX discards all number formatting (`_stringify_cell_value`, `:236-254`) — `$150,000` vs `150000` indistinguishable; a fringe rate stored as `0.475` renders `0.475`, not `47.5%`.
- XLSX merged headers listed as text (`:403-405`), not unmerged into the grid — a merged "TOTAL DIRECT COSTS" spanning B2:E2 lands in one cell with the rest blank.
- `reap_stuck_documents` matches `processing: False` (`document_tasks.py:1050`), so a document whose worker was killed mid-extraction — stuck at `processing: True, task_status: "extracting"` — is exactly the case it cannot reap.
- `pdf_page_count` returns `0` on error (`:723-725`), disabling `sparse_text` detection entirely for that document (`extraction_quality.py:49-51` returns `None` when `num_pages` is 0).

ViaJables:
Closes #812.

Five of the grab-bag, plus a docstring correction. Each is independent; grouped because they're all the same shape — the file shows one thing, the model reads another.

1. **XLSX percent formats** — Excel stores 47.5% as `0.475` and carries the `%` in the cell's display format, which extraction dropped: a fringe rate reached the model two orders of magnitude off. Percent cells now render as the sheet shows them. Currency and thousands separators are **deliberately not** reconstructed — `$150,000` and `150000` are the same quantity, and the bare number stays machine-readable.
2. **XLSX merged ranges** are spread across the cells they cover, so a "TOTAL DIRECT COSTS" merged over B2:E2 no longer leaves the columns beneath it unlabelled. The merge list in the extras is unchanged.
3. **`clean_markdown_nans`** only blanks cells that are *exactly* NaN. The blind `str.replace("NaN", "")` corrupted any word containing those letters — a PI's surname, "NaNoparticle".
4. **`DataExportNode`** ships non-tabular input as `.txt` with a warning naming the fix, instead of a prose blob labelled `.csv` that Excel opens without complaint.
5. **`reap_stuck_documents`** no longer requires `processing: false`, which excluded exactly the population it exists to rescue — a worker killed mid-extraction leaves that flag set forever. It now also matches in-progress documents whose `updated_at` predates twice the Celery hard time limit. The cutoff is deliberately built with a **naive** clock to match how `updated_at` is written; an aware UTC cutoff would skew by the server's offset and over-reap healthy in-flight documents on any host west of UTC.

Also: `pdf_page_count`'s `0`-on-error now says in its docstring and its log line that it means *unknown* and disables sparse-text detection for that document.

**Not in this PR:** the plain-text encoding papercut (bare `utf-8` open while `_read_text_with_fallback` sat unused in the same file) — already fixed by the latin-1 gate work in #828, which routed every plain-text read through one gated reader.

**Testing:** 7 new tests — NaN-only blanking with real words preserved, percent scaling (and pass-through for currency/non-numbers/bools), merged-header spanning through a real openpyxl workbook, the reaper's query shape plus a cutoff-vs-time-limit assertion, and Data Export's txt-with-warning / csv-unchanged split. 6 of 7 verified red against unfixed code. Full backend suite: 4456 passed, 0 failed. Ruff clean.

https://claude.ai/code/session_01CtRYNNEdrq6SPyPt8PjH1v


ViaJables:
Updated after review — three of my own fixes were wrong, two of them worse than the papercut:

- **Merged spreading applied to value cells.** A total of 485,000 merged across B10:E10 rendered four times; a model summing the row reads 1,940,000. Now only non-numeric anchors (labels) are spread — the case this was for — with a test pinning that a merged number appears exactly once.
- **The percent scale fired on literal percent signs.** `0"%"` and `0.0" %"` are cells that already hold 47.5, so they were being multiplied to 4750% — the same 100x error, backwards. Quoted and escaped `%` are excluded now, and the render carries 10 decimals instead of 4 (1e-7 was collapsing to "0%").
- **The `reap_stuck_documents` change is reverted.** It couldn't reach its target population (`raw_text` is blanked at extraction start, so the untouched `raw_text` guard still excluded stuck documents) and `updated_at` never advances, so the staleness cutoff was really age-since-upload. Details and what it actually needs are on #812, which stays open for that piece.

Title and changelog now say four, not five.

PR #857, opened 2026-09-04
fix(models): make vLLM structured output work, and make the Test button catch it

ViaJables:
Prod extractions and workflows were failing with `Native structured output is not supported by this model.` while chat kept working and dev was fine. Two tickets were filed for it.

## The bug

The vLLM path asks for schema-enforced output via pydantic-ai's `NativeOutput` (since v4.9.0). But `VLLMProvider.model_profile` only built an OpenAI-shaped profile for **bare** model names — a name containing a `/` fell through to OpenRouter's provider→family map, where `ModelProfile.supports_json_schema_output` defaults to `False`, and pydantic-ai refuses the request.

```
Qwen/Qwen3-32B              False   ← breaks
RedHatAI/Qwen3-32B-FP8      False   ← breaks
meta-llama/Llama-3.3-70B    False   ← breaks
qwen3-32b                   True    ← fine
```

Prod registered the model under its HuggingFace repo id, dev under a bare name. Same code, one string apart. JSON-schema output is a property of the **server** — vLLM enforces `response_format` through guided decoding for anything it serves — so the profile was answering for the wrong thing. `VLLMProvider` now declares the capability for every model it serves, keeping each family's own schema transformer (`qwen/…` still gets `InlineDefsJsonSchemaTransformer`).

The per-model **"supports structured output"** toggle also does something now. It was written by the model editor and read by nothing on the extraction path — only by the tuning candidate builder — so an admin switching it off, which is exactly the escape hatch this situation calls for, changed nothing about the request.

## Why nothing caught it

The admin **Test** button ran a single free-text completion. That proves reachability, credentials and the model name, and nothing else. It reported the model healthy for the entire outage, so the real cause stayed invisible until a user filed a ticket.

`diagnose_model` now makes a second, schema-constrained round trip using `use_native_structured_output()` — the *same* call the extraction engine makes, extracted so the two cannot drift. A diagnostic that reimplements the rule can pass on exactly the configuration a real run fails on. A model that chats but cannot return a structured object is now reported as failing: *"Connected, but structured output failed — chat will work, extraction and workflows will not."*

Two supporting changes from the same incident:

- **An unsupported output mode is classified.** It matched none of the transport categories, so it landed in `unknown` → "read the raw error" — for the one failure with a precise remedy. It now has a `capability` category naming both fixes.
- **The Endpoint step reports the URL actually dialed.** `VLLMProvider`/`OllamaProvider` append `/v1` to a stored endpoint; the OpenAI/InsightAI provider uses it verbatim. So one saved endpoint means different URLs depending on the protocol dropdown, and switching it silently changed the URL with nothing in the UI saying so — which is exactly how the first attempted workaround failed. Now: `http://host:8000 → dialing http://host:8000/v1`. This needed `unwrap_model_base_url()`: `base_url` resolves as a property on the `MeteredModel`/`RetryingModel` wrappers and returns `None` rather than falling through to the provider.

## Separate bug found on the way

Building a model to test the URL helper raised `TypeError: OpenAIChatModel.__init__() got an unexpected keyword argument 'openai_client'`. The external-OpenAI branch passed `openai_client=` to a constructor taking only `provider`/`profile`/`settings`, so **every model added through the "OpenAI" or "Custom" setup preset failed to build** — both presets save external + the openai protocol. No test built a model, so nothing noticed. Fixed by wrapping the client in an `OpenAIProvider`, with a regression test that builds one.

## Verification

- `4509 passed, 161 skipped` — full backend suite on this branch. `ruff check app/` clean.
- The new tests were confirmed capable of failing: reverting the two source files turns 9 red (`Qwen/Qwen3-32B`, `RedHatAI/…`, `meta-llama/…`, the toggle) while the bare-name cases stay green — the exact shape of the bug; reverting `system_diagnostics.py` alone turns 6 red.
- The pre-existing `TestVllmStructuredOutputMode` missed this because it uses the bare name `local-model` and mocks out `get_agent_model`, so the profile was never exercised.
- One existing test updated: `test_admin_routes.py`'s diagnostics mock now serves two round trips instead of one.

## Not in scope

Deferred as a follow-up, worth a ticket:

- Probe `multimodal` / `supports_pdf` / `context_window` the same way, letting the probe contradict the checkbox.
- Stop waiting for a button press — run the probe on model save, and/or a nightly re-test that fires a failure notification when a model that used to pass stops passing. That is the layer that catches prod drifting after a deploy, which is this incident's actual shape.

## Deploy note

Until this ships, prod is unblocked by setting the model's `api_protocol` to `ollama` (it appends `/v1` and skips the `NativeOutput` branch). Two costs while it sits there: thinking control goes silent — the ollama path sends `extra_body["think"]` instead of `chat_template_kwargs: {"enable_thinking": …}`, which vLLM ignores — and the context-window probe uses the wrong prober. Switch back to `vllm` after deploying this.

https://claude.ai/code/session_01PJ3BAa7MWo8CySegCumLEC


ViaJables:
Merged after review. One follow-up worth a test of its own:

- No test drives the real stack for a slash-named model — `test_llm_service` checks the profile, and the diagnostics tests mock `Agent`, so a future pydantic-ai change that moves the `supports_json_schema_output` check elsewhere would not be caught. One test that builds `get_agent_model("Qwen/Qwen3-32B", cfg)`, mocks the httpx transport with a canned `json_schema` chat completion, and asserts `Agent(model, output_type=NativeOutput(...)).run_sync(...)` does not raise `UserError` would pin the actual capability.
- `test_reading_base_url_off_the_wrapper_returns_nothing` asserts a pydantic-ai internal quirk; a pydantic-ai upgrade that fixes it turns the test red for no product reason.

Committed on the branch before merge: the structured-output failure summary now says "extraction and extraction-based workflow steps will not" — LLM prompt steps answer in free text and keep working.

PR #866, opened 2026-09-08
fix: generic container-engine support (podman / podman-compose)

ticklejw:
## Summary

`./setup.sh` and the compose stack assumed Docker specifically. On a stock RHEL host with podman + podman-compose, setup failed before pulling a single image, and a workaround install crash-looped celery. This PR makes the deploy path engine-agnostic; everything is verified live on RHEL 10 with podman 5.8.2 and podman-compose 1.5.0 (full stack healthy, celery at 0 restarts), and behavior under Docker is unchanged with the exception that a permanently unhealthy probe was fixed.

The goal is for this to be a step towards Kubernetes support, which I have ready to go but it depends on merging the changes in this PR in order to support generic OCI.

## Changes

- Fully qualify all image references (`docker.io/...`): podman's `short-name-mode=enforcing` can't prompt without a TTY, and RHEL's shortname config silently redirected `mongo` to a registry that 401s.
- Declare api/celery/frontend healthchecks at the compose level: OCI-format builds (podman's default) drop Dockerfile `HEALTHCHECK`. This also fixes a latent bug visible under Docker: the celery container inherited the API probe from the shared backend image and was permanently unhealthy — it now probes `celery -A celery_worker.celery_app status`.
- `run_celery.sh`: recognize podman and cgroup-v2 containers (`/run/.containerenv`, `libpod` in `/proc/1/cgroup`); previously PID 1 exited after starting workers (observed: 89 restarts).
- setup.sh/status.sh/upgrade.sh: discover containers via engine-CLI label filters and read state/health via `inspect` instead of Compose-v2-only `ps --format` fields; engine detection falls back docker→podman, compose detection docker compose→docker-compose→podman-compose; image checks accept both `project-svc` and `project_svc` names.
- CHANGELOG entry.

## Test Plan

- [x] Manually tested the affected feature(s) — full stack brought up healthy under podman-compose on RHEL 10; status.sh all-green; bootstrap + catalog seed ran; `/` and `/api/health` 200 through the published port

## Related Issues

None filed; happy to open one retroactively if preferred.

Also, I did not make any changes to the deploy/release docs to represent that Podman is now supported. I would be happy to do that but wanted to get a feel for the appetite for this first.

There was AI assistance here but I'm quite experienced with containers and I stand by this PR.


ViaJables:
Merged — thank you for this, and for the RHEL verification. Three things I changed on the branch before merging, so you know what landed:

1. **`compose.yaml` had two `healthcheck:` keys on the `celery` service.** `docker compose config` rejects the file outright (`mapping key "healthcheck" already defined`), which would have broken every Docker deploy; podman-compose's PyYAML keeps the last key silently, which is why the live test passed. `main` already had a compose-level `celery inspect ping` probe for celery (with the same reasoning in its comment), so I dropped the new block and kept that one. The CHANGELOG line no longer claims a latent Docker celery bug.
2. **`compose_container` now prefers a running container.** `ps -a | head -1` could return an exited or one-off `compose run` container (they list newest first), so `docker exec` targets in bootstrap / catalog update / re-ingest could hit a stopped container and fail with an exec error instead of the old "API container is not running". It tries `ps` first and falls back to `ps -a`. Same change in `status.sh`.
3. The CHANGELOG notes that `detect_deployment` moving to `ps -a` means a fully stopped stack now counts as an existing deployment rather than a fresh install — arguably better, but a visible change.

Two nits left as-is: `compose_project` ignores a top-level `name:` in compose.yaml or `COMPOSE_PROJECT_NAME` from a root `.env` (neither exists today), and `verify()` still hard-codes `vandalizer_${vol}` while `status.sh` uses `$(compose_project)_${vol}`.

PR #868, opened 2026-09-08
fix: SAML edge cases — x509certMulti metadata, forwarded proto, ACS redirect status

ticklejw:
## Summary

Three small fixes from bringing the SAML SP up in a deployment whose topology differs a bit from the common case: our IdP publishes separate signing and encryption certificates (`x509certMulti` rather than the flat `x509cert`), and TLS terminates at a proxy in front of the frontend nginx. Each fix is independent, small, and was verified against our live IdP; none changes behavior for setups that weren't hitting these paths.

## Changes

- `x509certMulti` metadata parsing: an IdP publishing distinct signing and encryption certs (standard Shibboleth) is rejected by parse-metadata as "missing … signing certificate" because only the flat `x509cert` key was read. Signing certs now come from `x509certMulti` (first used, alternates returned for key-rollover), and the 422 names the actually-missing fields.
- `X-Forwarded-Proto` clobbering: the frontend nginx terminates plain HTTP and stamped `$scheme` (always `http`) over the edge's truthful header, so AuthnRequests carried `http://` issuer/ACS URLs and strict-mode response validation rejected the IdP's POST-back. nginx now passes an upstream header through, falling back to `$scheme` only when none arrives.
- ACS 303: the ACS answers the IdP's POST and redirected 307 (method- and body-preserving), so a successful login re-POSTed the SAML envelope at `/` and got a 405 from the static frontend. Both ACS redirects are now 303.
- CHANGELOG entries for all three.

## Test Plan

- [x] Manually tested — full SAML round-trip (AuthnRequest → IdP → ACS → logged-in session) against our production Shibboleth IdP, with TLS terminated ahead of nginx; the metadata fix verified against that IdP's published metadata
- [x] Updated `CHANGELOG.md`


ViaJables (approved):
Reviewed against the code rather than the description, and the description holds up.

- **x509certMulti**: checked python3-saml's parser. It only flattens to `x509cert` when one cert serves both roles (or signing and encryption are identical); otherwise it emits `x509certMulti` with `signing`/`encryption` lists. Reading the signing list is the right fix.
- **nginx map**: safe. The frontend nginx only listens on :80 so TLS always terminates in front of it, and the only backend consumer of `X-Forwarded-Proto` is the SAML service. A direct client spoofing the header affects nothing but its own SAML exchange, which strict mode then rejects.
- **303 out of the ACS**: correct HTTP semantics for a POST-answering handler.

To get this merged without a round-trip I pushed two commits to your branch:

1. A merge of `main` resolving the `CHANGELOG.md` conflict with the v4.12.0 stamp (your three entries now sit under `[Unreleased] > Fixed`).
2. `eee58f3b`, three small things from the review:
   - The ACS *failure* redirect still defaulted to `/login`, which forwards to `/landing` without its query string, so `?error=saml_failed` was never shown. Defaulted it to `/landing`, matching the OAuth callback. You had spotted this in #869's comment; this puts it where the line lives.
   - `tests/test_saml_metadata_parse.py`: five tests that push real Shibboleth-shaped metadata (distinct signing/encryption KeyDescriptors, a key-rollover pair, an SSO-less descriptor) through the real parser. Three of the five fail against the pre-PR endpoint. The parse-metadata endpoint had no coverage at all before.
   - Wording: the nginx comment and changelog said the header drove cookie `Secure` flags (those come from `settings.use_secure_cookies`) and that the map could yield an empty value, which it never does.

One thing left as-is: `idp_x509_cert_alternates` is returned by the endpoint but the frontend type and AuthPanel ignore it, so the rollover hint is only visible in the raw response. Fine to leave; worth surfacing in a follow-up if you want admins to see it.

Thanks for the careful writeup. The live-IdP verification on each item made this easy to trust.

PR #869, opened 2026-09-08
feat: per-provider JIT-provisioning toggle for SSO

ticklejw:
## Summary

Any identity the IdP asserts is auto-created on first SSO login, with no way to require accounts to exist beforehand. Each provider (Azure OAuth and SAML) now carries a "Create accounts on first sign-in (JIT provisioning)" checkbox. **Default on — and providers saved before the field existed read as on — so no existing deployment changes behavior.** Based on the saml-fixes PR (#868 ); review the last two commits.

## Changes

- `jit_provisioning: bool = True` on the provider record (plain bool with a default on purpose: `model_dump(exclude_none=True)` then always persists it, so the full-replace provider update handler can't silently drop it — the same field-loss class #852 fixed for model config).
- `resolve_oauth_user` / `resolve_saml_user` raise `JitProvisioningDisabled` instead of creating when the flag is off; existing users are unaffected (the flag gates creation only — lookup, email-based account linking, and attribute sync unchanged).
- Both SSO routes catch it, audit-log `user.login_denied` with the asserted ID, and redirect to `/landing?error=sso_user_not_provisioned` (303 from the ACS, which answers a POST); the landing page renders a specific "contact your administrator" message for that code.
- Admin UI checkbox on both add- and edit-provider forms, prefill treats an absent stored key as on.
- Tests: 3 backend (deny-unknown on both provider types; allow-existing with the flag off; the untouched existing tests are the default-on backward-compat check) and 3 frontend (add payload, stored-false surviving the full-replace update, legacy prefill).

## Test Plan

- [x] Manually tested — live cluster with a Shibboleth IdP: unknown identity denied with the message and audit entry; existing account logs in unchanged; re-enabling restores auto-create
- [x] Backend `tests/test_auth_service.py` 31/31, frontend AuthPanel tests 18/18, `tsc -b` clean
- [x] Updated `CHANGELOG.md`

## Related Issues

Closes #


ViaJables (approved):
Reviewed the two commits on top of #868. Approving.

- **Gate placement**: both resolvers do the three lookups (upn, email, user_id==email) first and raise only on the create path, so existing users, email-based linking, and attribute sync are untouched, as described.
- **Persistence**: add and update both `model_dump(exclude_none=True)`, so the bool always lands in the stored dict; `_sanitize_providers` copies the dict wholesale, so the admin prefill sees the stored value. Both SSO routes read the flag off the stored provider with a default of `True`, so pre-existing providers behave exactly as before.
- **Error surface**: the `/landing` route's `validateSearch` passes `error` through as a string and the page renders the specific message.
- **Tests**: deny-unknown on both provider types, allow-existing with the flag off, and no `insert` on denial. Adequate for this size.

Pushed a merge commit resolving `CHANGELOG.md` against the v4.12.0 stamp and picking up the updated #868 tip (which now also defaults the pre-existing `saml_failed` redirect to `/landing`, the thing your comment here identified). Your entry sits under `[Unreleased] > Added`. Backend auth/SAML tests: 48 passed locally.

Will merge #868 first, then this, once CI is green on the new pushes.

PR #874, opened 2026-09-09
fix(llm): output cut-off warning names user remedies; admin remedy shown to admins only

ViaJables:
## Support ticket

When a model response hits its output token cap, the warning stored on the workflow step said: *"Raise 'Response reserve (output tokens)' for this model under Admin → System Config → Models, or split this step into smaller steps."* Regular users can't open the Admin area, so the message pointed them at a page that doesn't exist for them. The same instruction appeared on the validation-run detail note for a truncated answer.

## Change

- **Backend** (`describe_truncation`): the stored warning now names only remedies any user can act on — ask for a tighter output format, split the step into smaller steps, or run it per document instead of across the whole set. It still opens with the cap ("stopped at its 8,192-token output limit") so the cause is unchanged. Exports and the API return this text, so it reads correctly for a non-admin there too.
- **Frontend** (`utils/truncationWarning.ts`): `withAdminRemedy(warning, isAdmin)` appends *"As an admin, you can also raise 'Response reserve (output tokens)' … under Admin → System Config → Models."* when the viewer is an admin and the warning is the cut-off one (keyed on the shared `Output was cut off` prefix, which the backend exports as `TRUNCATION_WARNING_PREFIX`). Applied on the workflow editor's last-run output, the step-warnings panel on the output card, and the step test result. `TruncationNote` in the KB validation run tab gates its admin sentence the same way and gives non-admins the fix they can make.

## Tests

- Backend: new `test_describe_truncation_names_only_user_remedies` (asserts no "Admin" / "Response reserve" / "System Config" in the stored text, and that each user remedy is named). `pytest tests/test_llm_service.py -k truncation`: 7 passed.
- Frontend: `truncationWarning.test.ts` — regular user gets the stored text unchanged, admin gets the appended remedy, no double-append, matches when joined onto another step's warning, ignores unrelated warnings. 5 passed. `tsc --noEmit` clean; eslint on the touched files shows only two pre-existing warnings.


ViaJables:
Reviewed against the code. All claims verified: the stored warning carries only user remedies (`describe_truncation`, test asserts absence of the admin text), the three editor render sites plus the validation-run note gate on `user?.is_admin` like the rest of the app, no other consumer of `steps_output` renders the warning, and the only other assertion (`"cut off" in out["warning"]`) still holds.

Nits, not changed: `useIsAdmin()` in `utils/truncationWarning.ts` is a hook in a utils file used only by `TruncationNote`, while the editor sites use `useAuth()` directly; one pattern would be cleaner. The non-admin sentence in the validation note ("Shorter or more focused test queries fit within the limit") is a weak fit since the cap is on the answer, not the query.

Merged with main on the branch (CHANGELOG only).

PR #875, opened 2026-09-09
fix(kb-validation): drop a stale override model instead of grading empty answers

ViaJables:
## Support ticket

2 CFR 200 KB validation exports show blank `actual_answer` on every query in four runs since 1 September, beside populated expected answers and an empty `error` column. The judge fails each one and the KB's score fell from 76 to the 40s. Another KB still populated answers on 6 September.

## Cause (confirmed against production)

Autovalidate "apply" stored the winning answer model by name on the KB's `rag_config_override`: `qwen/qwen3.6-27b`. Between the 24 August run and the 1 September run the admin renamed that model in System Config to `qwen/qwen3.8-27b` (production's model list today has only `openai/gpt-oss-120b` and `qwen/qwen3.8-27b`; the judge model on the exports changed at the same time). The override name then matched nothing:

- `_build_agent_model` still builds an agent for an unknown name, with no API key and the global endpoint, so every answer call failed.
- `_generate_kb_answer` swallowed the exception into `""`, and the judge graded the empty string as a FAIL.
- The per-query row never wrote an `error`, so the export's `error` column read as "no error".
- Retrieval doesn't need the model, and the no-KB baseline uses the user's model, so both kept working. Only this KB has an override that pins a model.

## Fix

- **Resolution drops a stale override model.** `_resolve_rag_config` checks the override's `model` against System Config; a name that no longer exists is dropped (a tag resolves to its name; a config outage keeps the name). The tuned retrieval knobs still apply and the caller's model answers. Live chat on such a KB also stops sending its query-rewrite and rerank calls to the dead model.
- **The run says which model answered.** `run_kb_validation` records `answer_model` and an `answer_model_fallback` note on the result, in `model_settings`, and in the export's run metadata. The validation header shows an amber "tuned model unavailable · answered by …" line.
- **A generation failure is no longer an empty answer.** `_generate_kb_answer` propagates; the per-query handler makes it a SKIPPED row with no score (the judge-unavailable contract from #a462a33b), and the row carries the error so the export's `error` column is populated.
- **Baseline uses the same answer model as the KB answer** when an override names one: the resolved config is pinned for the whole run, so lift measures the KB rather than a model swap.

## After deploy

The 2 CFR 200 KB's override still names `qwen/qwen3.6-27b`. Validation will run on the user's model and say so in the header. To get the tuned model back, re-run Autovalidate on that KB (or revert the optimization).

## Tests

- New: stale override model dropped / existing model kept / tag resolved / config outage keeps the name; `run_kb_validation` records the fallback note and pins the config; no note when the model exists; export carries `answer_model`, `answer_model_fallback`, and the row `error`.
- Changed: `test_generate_kb_answer_swallows_agent_errors` → `..._propagates_agent_errors`; the per-query failure test asserts the row's `error`.
- Red check: all five new/changed service tests fail against the previous service code.
- 319 passed across the KB validation, export, optimizer, retrieval, routes, and regression-suite modules; `tsc -b` and eslint clean.

https://claude.ai/code/session_01BE6Mv19eNWBDxXJHmvCyjD


ViaJables:
Reviewed against the code. Verified: stale override detection is an exact-name lookup then a tag scan; the tuned retrieval knobs survive the drop; `answer_model` / `answer_model_fallback` reach the result, `model_settings`, the export's run metadata and the panel header; a generation failure is now a SKIPPED row with `error` instead of a graded empty string; chat retrieval on such a KB stops calling the dead model too.

One fix pushed on the branch: the fallback note fired on `answer_cfg.model != override_model`, which is also true when the override held a **tag** ("Strong") that resolved fine. The run would then claim "tuned model unavailable" while using the tuned model. The note now fires only when `_configured_model_name(override)` is `None`, with a test (red without the fix).

Behavior change worth knowing: the resolved config is now pinned for the baseline too, so a KB whose override model *does* exist has its no-KB baseline answered by the override model rather than the user's model. That is the right comparison, but it shifts lift numbers on such KBs.

Not changed: the SKIPPED row's `error` always says "answer generation failed: …" even when the judge or retrieval raised.

Merged with main on the branch (CHANGELOG only).

PR #876, opened 2026-09-09
feat(kb): auto-generated test queries get stable IDs, source, and provenance notes

ViaJables:
## Support ticket

> Please make autogenerated test queries have these fields so it would be standardized with imported sets: Question, Expected Answer, Category, Source or Section, Notes, and ID. Preserve imported IDs as provided, and assign autogenerated questions a stable, human-readable ID such as [KB-prefix]-AUTO-Q001.

## What changes

**ID.** New `app/services/kb_test_query_ids.py`. Every generated question gets `<KB-prefix>-AUTO-Q<nnn>`:
- Prefix from the KB title: one word keeps up to six letters (`FCOI`), several words take initials of the significant ones (`Export Control Regulations` → `ECR`, `FCOI Knowledge Base` → `FCOI`); an empty/punctuation title falls back to `KB` + four uuid chars.
- Number continues from the highest `-AUTO-Q` already on the KB (any prefix, so a renamed KB does not restart the sequence), and the allocator skips anything already taken, so a second generation never reuses an ID.
- **Imported IDs are never touched.** The importer's upsert-by-ID behaviour is unchanged; a row whose ID matches an auto ID updates that row in place, as designed.
- **Backfill.** Auto-generated queries from before this change get IDs on the next `GET /test-queries`, oldest first, continuing the sequence. Deterministic, so two concurrent reads assign the same IDs to the same rows; a backfill failure is logged and does not break listing.

**Source or Section.** When the model's `expected_source_labels` name no real source (they are filtered today), the Source column is now filled from the source names of the chunks it cited — real names by construction — so it is never blank for a chunk-grounded question.

**Notes.** Provenance line: `Auto-generated 2026-09-09 from Award Letter (standard coverage, model qwen3.8).`

Question, Expected Answer and Category were already populated; all three generation paths (sync route, Celery task, optimizer's first-run autogen) go through `KBQuestionGenerator.generate`, so they all get this.

**UI.** The Test queries tab shows the ID as a chip at the front of each card (replacing the small `· ID:` meta text) and gains a search box narrowing by ID, question, expected answer, category, source labels or notes. The generate modal says what columns and ID form to expect.

**Exports.** Validation-run detail rows now record `notes` alongside `external_id`, and the CSV/XLSX/JSON export gains a `notes` column (after `expected_sources`), with the live query as fallback for older runs.

## Tests

- `tests/test_kb_test_query_ids.py` (new): prefix derivation (single/multi-word, stopwords, cap, fallbacks, non-string title), numbering (starts at 1, continues past max, old prefix counts, collision skip, no repeats), notes text, backfill (oldest first, imports and hand-written rows untouched, no-op when complete).
- `tests/test_kb_question_generator.py`: end-to-end asserts `ECR-AUTO-Q001` + notes; new tests for continuing numbering with imported IDs present and for the cited-chunk source fallback.
- `tests/test_kb_validation_export.py`: notes column carried.
- `KBTestQueriesTab.test.tsx`: `matchesSearch` over each column; ID chip renders; typing an ID narrows the visible rows; empty-state names the term.

Backend: 275 passed across the touched/related files (`test_kb_test_query_ids`, `test_kb_question_generator`, `test_kb_validation_export`, `test_kb_validation_service`, `test_knowledge_routes`, `test_kb_optimizer`, `test_kb_validation_tasks`, `test_quality_service`), ruff clean. Frontend: 12 passed in the tab test file, `tsc --noEmit` clean, eslint clean on touched files.


ViaJables:
Reviewed against the code. Verified: prefix derivation, continuation past the highest `-AUTO-Q`, imports never touched, backfill on GET oldest-first, all three generation paths through `generate`, Source column from cited chunks, UI chip + search tolerant of older rows lacking `external_id` / `notes`.

Fixes pushed on the branch:
- **Concurrent allocation.** No unique index on `external_id` and no per-KB lock, so two generations (route, Celery task, optimizer autogen) or a generation racing a backfill could read the same IDs and both mint the same number. Persisted rows now `reserve()` each ID against the KB before insert (`find_one` re-check, skip on hit); a preview still allocates in memory; the backfill excludes the row it is about to write so a concurrent backfill stays deterministic. Tests are red without it.
- The allocator's `while` skip loop could never iterate (any taken `prefix-AUTO-Q{n}` already raised the counter past `n`), and its test used a trailing-space ID that never collided. Both removed; the loop's job is now `reserve`.
- `kb_id_prefix("2 CFR 200")` produced `2C2`; numerals now yield to words (`CFR`). Titles with only numerals keep them.

Note on "stable": IDs are assigned once and never reused, not derived from the question text, so a regenerated set gets fresh numbers. That matches the ticket but the changelog's "stable" could be read the other way.

Not changed: recorded `notes` on validation-run detail rows has no test beyond the live-query fallback; the backfill writes on a GET.

Merged with main on the branch (CHANGELOG).

PR #877, opened 2026-09-09
feat(kb-validation): run a validation over selected test queries

ViaJables:
## Support ticket

In Validation › Test Queries, selecting questions only offered deletion. The evaluator wants a "Run selected" option to smoke-test a few questions and get the JSON from run history for just those.

## What changed

- **"Run selected (n)"** appears beside "Delete selected" on the Test Queries tab. It starts a judge-mode validation over the chosen queries; the panel owns the run, so the result shows on the Run tab (with a note: "Smoke test over n of N test queries…") and in History, where the row is labeled "selected n/N" and its JSON/CSV/Excel export holds only those questions. The export's run metadata carries `query_selection: {selected, total}`.
- **API:** `POST /api/knowledge/{uuid}/validate` accepts `query_uuids` (deduplicated; 400 when empty, malformed, or when none belong to this KB). The Celery task passes it through.
- **A smoke test is not the KB's quality score.** The run is persisted with `source="smoke_test"`, and:
  - `persist_validation_run` skips the quality-metadata update for it (new `source` parameter);
  - both "latest run" readers in `quality_service` (the one that sets `VerifiedItemMetadata`, and `get_latest_validation`, which regression detection compares to) filter it out via one shared `NOT_SMOKE_TEST` clause;
  - the KB trust summary, the `latest` export pick, and the first-run judge-variance check leave it out;
  - the validation header keeps the latest full run.

## Notes

- Judge mode only (no baseline) for smoke tests, so the cost is what the handful of questions costs. The Run tab's mode selector still governs full runs.
- Touches the `run_kb_validation` result/persist block that #875 also edits; expect a small conflict if both land.

## Tests

- New `tests/test_kb_validation_run_selected.py`: subset run judges only the selected queries, records `query_selection`, persists as a smoke test; a full run is untagged; no matching query is a `ValueError`; a smoke-test persist skips quality metadata; both latest-run readers carry the filter; export metadata carries the selection.
- Route tests (`TestValidateSelectedQueries`): uuids reach the task deduplicated after an ownership check; a full run passes none; empty/malformed/foreign selections are 400.
- Tab tests: "Run selected" calls back with exactly the selected uuids, is absent when not offered, and is disabled while a run is in flight.
- 218 backend tests pass across the touched suites; `tsc -b` and eslint clean; the two vitest files pass (15 tests; on this laptop the first test of each file needs a longer than 5s timeout for cold start, which also affects an untouched history-tab test).

https://claude.ai/code/session_01BE6Mv19eNWBDxXJHmvCyjD


ViaJables:
Reviewed against the code. Verified: selection by `KBTestQuery.uuid` (no dependency on #876), 400 on empty/malformed/unowned selection, Celery task passes `query_uuids` through with the retry decorator untouched, `persist_validation_run` skips quality metadata for `smoke_test`, both "latest run" readers, the trust summary, the `latest` export and the first-run variance prior share `NOT_SMOKE_TEST`, history rows labelled "selected n/N", per-run export holds only the selected questions.

Fixes pushed on the branch:
- **Admin quality aggregates still counted smoke tests.** `get_quality_summary`, `get_quality_timeline`, `get_quality_by_model`, `get_quality_items` (trend arrow) and `get_quality_item_detail` (per-model averages) read every run, so a three-query smoke test scoring 100 could flip an item's trend or enter a fleet average, contradicting the "not the quality score" guarantee. All five now filter on `NOT_SMOKE_TEST`; history is left as is. Source-inspection test added alongside the existing one (red without the fix).
- A smoke test that is the first run on a KB no longer runs `_sample_judge_variance_detailed`; its variance was never read.
- Renamed the tab test that promised a disabled assertion it did not make.

Not changed: sparkline bars for smoke tests are full-height in the same colour (only the hover title says "smoke test"); no upper bound on `query_uuids` length, and `query_selection.selected` reports matched rather than requested count.

Merged with #876's branch and main; conflicts were adjacent additions in `kb_validation_export.py`, `kb_validation_service.py`, `api/knowledge.ts`, the lucide import in `KBTestQueriesTab.tsx`, and CHANGELOG, all resolved by keeping both sides.

PR #878, opened 2026-09-09
docs(certification): Extraction is optional and task-dependent, not the mandatory first step

ViaJables:
## Support ticket

> Many certificate examples begin with an Extraction node, which kind of imply extraction should always come first. For short or context heavy documents, sending the whole file directly to a Prompt can be more helpful as we discovered from previous experience. I suggest the training clarify that Extraction is optional and task-dependent.

## What the course said

- Module 2 "Common processes": *"every workflow starts with extraction"*.
- Module 3 "The extract-reason-deliver pattern": every phase-1 example is the Extraction task; the knowledge check's correct answer read *"Always start by getting clean data out of the document."*
- Module 5: lesson titled *"Design principle: extract first, reason second"*.
- Module 6: Extract → Analyze → Render mapped one-to-one to Extraction → Prompt → Output node.
- The Module 5 and 6 labs script "Step 1 — Extraction" with no reason given.

## What it says now

The engine already supports the ticket's design: the first step after the Document trigger receives the workflow's documents automatically (`_resolve_input_sources` swaps `step_input` for `workflow_documents`), and any later step can pick **Workflow Documents** or **Selected Document**. The lessons now teach that.

- **Module 2** — the AI-suitability table gains *"read a short or context-heavy document and answer a question about it → Prompt, given the whole document"*; the "every workflow starts with extraction" line becomes "starts by *reading* the document", with the short-document case spelled out.
- **Module 3** — the task-type mapping gains the Prompt-with-document row. The extract-reason-deliver lesson gets a paragraph: *"The Extract phase is optional, and task-dependent"* — when Extraction earns its place (fixed fields, long documents, checkable values) and when to send the whole file to the Prompt, and how. The **knowledge check is rewritten** to ask when to skip Extraction (the old right answer said never to). The split-or-combine rule, the subaward worked example, and the glossary each gain the short-document case. Module objective updated.
- **Module 5** — the Prompt-node lesson says a Prompt needs no Extraction in front of it and names the input-source control; the principle lesson is retitled **"separate reading from reasoning"** and explains that the principle is separation of concerns, not "Extraction goes first".
- **Module 6** — the end-to-end lesson and its knowledge-check explanation note that phases 1 and 2 collapse into one Prompt for a short source.
- **Labs** (`exercises.json`) — the multi-step and output-delivery labs each gain one closing note saying *why* Extraction is first for that document (long, fixed terms) and that it is optional otherwise. **Lab requirements, validators and star criteria are unchanged** — the multi-step validator still requires Extraction + Prompt + Formatter, and the note says so is the point of that lab.

The `extract-reason-deliver` diagram is unchanged: the pattern is still right, the nuance is what "Extract" means.

## Verification

- `exercises.json` still parses; `test_certification_extraction_fields.py` + `test_certification_governance.py`: 13 passed.
- `tsc --noEmit` clean; eslint clean on `modules.ts`.
- Prose-only change; no code paths touched.


ViaJables:
Reviewed against the code. Content direction verified: the engine supports Prompt-first (`_resolve_input_sources`), the old "extract first" text is fully gone, lab validators and star criteria are untouched, and `certification-data` is not under the seeds VERSION gate.

One fix pushed on the branch: the new text justified the lab's Extraction step by document length ("the agreement is long", "a 40-page agreement"), but the lab's `subaward-agreement.pdf` is two pages, and the same PR teaches that a two-page letter should skip Extraction. The two sentences now justify it by what is true (fixed fields the lab checks one by one; practising a chained pipeline). Also corrected "Modules 2, 3, 5 and 6" to 7 (the end-to-end lesson is `output_delivery`, Module 7).

Merged with main on the branch (CHANGELOG).

Issue #633, opened 2026-08-12
[Bug] OCR treats a temporary outage as an empty document


arhyneRWU:
## Description

`ocr_extract_text_from_pdf` makes 3 in-process attempts with `2 ** attempt` backoff — a total wait budget of about **3 seconds** — then logs a warning and returns `""` (`backend/app/services/document_readers.py:152-178`). Callers cannot distinguish that empty string from a document that genuinely has no text, so a temporary OCR outage is recorded as an empty or degraded document and recovery is manual.

The 3-second budget is the core of it. Any OCR outage lasting longer than a few seconds — a busy Docling-Serve instance, a cloud 429, a service restart during deploy, a GPU held by another workload — exhausts all three attempts before the service has had any realistic chance to recover.

## Steps to Reproduce

1. Configure an OCR service in **Admin → System Config → Endpoints**.
2. Make it briefly unavailable — stop it, saturate it, or point at an endpoint returning `503`.
3. Upload a scanned PDF with no text layer.
4. Wait for processing to finish.

## Expected Behavior

A transient OCR failure is retried over a window comparable to a real outage, and if it ultimately fails the document is recorded as **failed for a retryable reason** — distinguishable from a document that legitimately contains no text.

## Actual Behavior

All three attempts are exhausted in ~3 seconds and `""` is returned. Two outcomes follow, depending on the PDF:

- **No text layer at all** — both PDF paths check `len(ocr_text.strip()) >= MIN_PDF_TEXT_LENGTH` (100 chars) and fall through to PyMuPDF, which returns nothing, so the document ends up `task_status: "error"` asking the user to retry. Correct-ish, but the user is told to retry a *permanent-looking* failure that was actually a 3-second outage.
- **A thin text layer over 100 chars** — the degraded PyMuPDF output is accepted **silently** as the document's text. No error, no signal that OCR never ran, and the document is then used for chat and extraction as though it were fully read.

## Environment

- **Vandalizer version/commit**: `main` as of this filing; the retry loop is unchanged in current `main`.
- Reproduced against a self-hosted OCR service; the defect is in the retry policy, not any particular provider.

## Additional Context

[...]

PR #887, opened 2026-09-11
fix(extraction): refuse a garbled PDF text layer the classifier rejected, and let a retry force OCR

### LINKED ISSUE #858: [Bug] A garbled PDF text layer is stored and warned about instead of being re-OCR'd as a scan
## Description

When a PDF's text layer is corrupt — subset fonts with no usable `ToUnicode` CMap, so extraction yields raw glyph IDs rather than characters — Vandalizer stores the mojibake as the document's text and then warns the user about it in chat. The correct response is to treat that text layer as absent and OCR the document as if it were a scan: rasterize the pages and read them with the vision path, which is exactly what the existing OCR route already does with the raw PDF.

Every acceptance decision in the PDF reader is a **length** check. `nonletter_ratio` — the function that already knows this text is garbage — is never consulted until after the reader has committed, so a document that scores 0.985 against a 0.25 threshold is accepted, stored, embedded into ChromaDB, and only then labelled unreliable.

Three consequences:

1. **The recovery advice we print does not work.** The chat notice says *"Try Retry extraction on the document."* `POST /documents/{uuid}/retry-extraction` re-dispatches the identical chain with no force-OCR flag (`routers/documents.py:136-172`), so the reader makes the same decision and stores the same garbage. The instruction is a no-op for the failure mode it is printed for.
2. **It does not require an OCR outage.** `_local_markdown_extract_from_pdf` gates on `pdf_type == "text_based"`, classifier confidence, an empty `pages_needing_ocr`, and `len(text) >= MIN_PDF_TEXT_LENGTH` (`document_readers.py:982`). A CID-broken PDF that `pdf-inspector` classifies as text-based takes the fast path and skips OCR entirely, with a perfectly healthy OCR service. Nothing in that path can notice the text is unreadable.
3. **The garbage reaches retrieval.** The document is chunked and embedded like any other, so it competes for top-k in KB and chat retrieval as well as being directly selectable.

## Steps to Reproduce

1. Upload a PDF whose embedded fonts defeat text extraction (subset fonts, no `ToUnicode` CMap). Ours was a 45-page policy manual saved from a web viewer.
2. Arrange for the reader to reach the PyMuPDF fallback — any of: OCR unconfigured, OCR returning a `PERMANENT_STATUS_CODES` failure, OCR returning under 100 characters, or the `pdf-inspector` fast path claiming the file.
3. Wait for extraction to finish. The document lands green — no error, `task_status` is not `error`.
4. Select it in chat and ask anything.

Observed on our instance: `extraction_nonletter_ratio: 0.9849409849409849`, 3,298 stored characters over 45 pages, 43 exact (non-interpolated) page markers, `ingest_error: null`. PyMuPDF on the same file yields 3,300 characters beginning:

```
'\x00\x01\x02\x03\x04\x03\x05\x06\x08\x08\x08\x00\t\x01\x04\x05\n\x0b\t\x05\x06\x08\x0c…'
```

`pdf_inspector.classify_pdf` on this file returns `image_based`, confidence 0.80, all 45 pages in `pages_needing_ocr` — so the classifier had it right and the fast path correctly declined. The document still ended up stored from PyMuPDF.

## Expected Behavior

A text layer that scores above `extraction_max_nonletter_ratio` is not text. Treat it as an empty extraction and take the scan route: send the PDF to OCR so the pages are rasterized and read, and prefer that result over the local text layer. The plumbing already exists — `ocr_client._post_pdf` posts the raw PDF and the service rasterizes it, so this is a routing decision, not new rasterization code.

Concretely:

- Gate on the classifier, not on a ratio. `_local_markdown_extract_from_pdf` already calls `pdf_inspector.classify_pdf` and discards the result; classify once at the top of `_read_pdf_text_and_markers` and carry it. When the classifier returns `image_based` — "the text layer on every page is untrustworthy, OCR this" — and OCR cannot run, do **not** fall back to the PyMuPDF text layer the classifier just rejected. Measurement below shows why `nonletter_ratio` cannot be this gate at any threshold.
- If OCR cannot be reached, this is an *unreadable document we could not recover*, not a successful extraction — it belongs in the same error state as the empty-text branch, with the honest message, rather than green with a warning.
- Give `retry-extraction` a way to force the OCR path, so the advice in the notice is actionable.

## Actual Behavior

The mojibake is stored as `raw_text`, the document processes green, it is chunked into ChromaDB, and the only signal is a chat-time notice:

> Text extracted poorly from 1 selected document(s): …. Most of the stored text is unreadable, so answers about these documents are likely to be unreliable or wrong. Try "Retry extraction" on the document, or re-upload it (e.g. as a scanned/printed copy) so OCR can produce clean text.

— whose suggested remedy deterministically reproduces the same result.

## Screenshots

N/A — backend behaviour. The stored state is quoted above.

## Environment

- **Browser**: N/A (backend ingestion path)
- **OS**: Ubuntu 22.04, Dockerized stack
- **Vandalizer version/commit**: code references are against `main@b32f34f7`. The instance where this was observed runs an older image whose `document_readers.py` predates the #633 `OcrUnavailableError` retry path, which is why our particular OCR failure (HTTP 503, transient) degraded to PyMuPDF instead of retrying. On `main` that specific 503 now retries correctly — but the missing quality gate described here is present on `main@b32f34f7` and is reached by every other route to the PyMuPDF fallback, and by the fast path regardless of OCR health.

## Measured impact

Measurements taken to size the fix, anonymized. Scored with the real `app.utils.extraction_quality.nonletter_ratio` over PyMuPDF `page.get_text()`; classification via `pdf_inspector` 0.2.6 `detect_pdf`. 113 real PDFs on a research-administration workstation plus ~40 constructed pages.

### A garbled text layer is not rare, and it is not always obvious

Two real documents had a garbled text layer (mojibake from a broken font encoding — subset fonts with a missing or invalid `ToUnicode` CMap). Both were produced by ordinary tools a research office meets every week:

| Document | Pages | What the text layer contains | `nonletter_ratio` | `pdf_inspector` |
|---|---|---|---|---|
| Institutional purchasing-policy manual, saved from a web page | 45 | Glyph IDs in the control range: `\x00\x01\x02\x03\x04…` | **0.985** | `image_based`, 45/45 pages need OCR |
| Federal grant application preview, exported from a sponsor portal | 102 | Mixed: `\x1e-*\x1e/\x04/(\x1e"\x1f)#(%)` alongside `HHIJHKLHHH MNOPQ RHHS` | **0.509** | `image_based`, 102/102 pages need OCR |
| Partly-scanned proposal (legitimate) | 79 | Real text | 0.0004 | `mixed`, 11/79 |
| ~110 other proposals, budgets, biosketches, subawards, scans (legitimate) | — | Real text, or empty | ≤ 0.006 | `text_based` / `scanned` |

The second document is the case that matters. It is a complete 102-page application whose text is garbage end to end, and only half its characters look wrong to the heuristic: **20% of the garbage is `isalnum()`** and **29% lands on characters in `_EXPECTED_PUNCTUATION`**. Per page, 61 of 102 score below 0.6 and 28 score below 0.25. Today it clears every acceptance gate, is chunked into retrieval, and is sent to the model with an advisory notice — and the notice's suggested remedy re-runs the same decision.

### No ratio threshold separates garbled from legitimate

| Content | `nonletter_ratio` | Why |
|---|---|---|
| Form page ruled with `═` / `║` box-drawing (legitimate) | 0.608 | box-drawing is not `isalnum()` and not in the expected set |
| Axis-label figure using U+2212 MINUS (legitimate) | 0.630 | U+2212 is not in the expected set |
| Vocalized Arabic / pointed Hebrew / Bengali prose (legitimate) | 0.35 – 0.45 | combining marks (Mn/Mc) are not `isalnum()` |
| Bilingual English/Hindi abstract (legitimate) | 0.226 | already at the edge of today's 0.25 advisory threshold |
| **Real 102-page garbled application** | **0.509** | half its garbage is `isalnum()` or expected punctuation |
| Garbled layer whose glyph IDs land in printable ASCII or CJK (constructed) | **0.000** | indistinguishable from clean text |
| Garbled layer whose glyph IDs land in the control range, PUA, or geometric shapes (constructed; the 45-page case above) | 1.000 | the only kind the ratio reliably catches |

The two distributions overlap between roughly 0.16 and 0.63. A reject threshold of 0.6 would have rejected the legitimate ruled form and passed the real 102-page garbled application. Nothing real lies between 0.63 and 0.98, so any threshold below ~0.85 buys only false positives, and a threshold above it catches only the control-range case.

### The classifier already separates them

| `pdf_inspector` signal | Fires on |
|---|---|
| `has_encoding_issues` | Neither garbled document; **does** fire on the legitimate partly-scanned proposal. Not usable. |
| **`pdf_type == "image_based"`** | **Exactly the two garbled documents; 0 of 89 legitimate files.** |
| `mixed` | The partly-scanned proposal only — separable. |

Constructed false-positive candidates for an `image_based` gate: an Acrobat-style searchable scan (full-page images with an invisible, legitimate text layer, common for signed subawards) classifies `text_based` at confidence 1.0 and is untouched. A figure-heavy document with short real captions does classify `image_based`; with OCR unavailable it would land in the error state instead of storing ~40 characters per page of captions — the same outcome a plain scan gets today, and the classifier is not wrong about it.

Caveats: n = 2 on real garbled documents. The `nonletter_ratio` false positives above (U+2212, box-drawing, checkbox glyphs, combining marks) are worth fixing for the *advisory* warning's sake, but that is #834 territory and separate from this gate.

## Additional Context

Where the signal already exists and is not used — verified against `main@b32f34f7`:

- `utils/extraction_quality.py:18` — `nonletter_ratio`, with a docstring that describes precisely this case ("a CID-mangled text layer measures ~0.5").
- `config.py:199` — `extraction_max_nonletter_ratio = 0.25`.
- `tasks/document_tasks.py:390` — the only place the ratio is computed, *after* the reader has returned and the text is about to be stored.
- `services/document_service.py:77` — `is_extraction_low_quality`, reading the stored value back for display.
- `services/document_readers.py` — no reference to `nonletter_ratio` anywhere. Acceptance is `len(text.strip()) >= MIN_PDF_TEXT_LENGTH` at `:982` (fast path) and `:1046` (OCR), and unconditional at the PyMuPDF fallback.

Related but distinct: #834 (4) notes that three "is this text?" definitions now coexist and should converge. This issue is about a fourth gap — none of the three is consulted at the point where the PDF reader decides which extraction to keep.

Adjacent prior work: #633 (an OCR outage must retry rather than be recorded as an unreadable document) fixed the *transient* case. This is the complement: when we do end up with a text layer we cannot read, re-OCR it rather than store it.

arhyneRWU:
## Summary

A PDF whose text layer is glyph IDs (subset fonts with no usable ToUnicode map, common in web-printed and sponsor-portal exports) no longer lands green with mojibake stored and chunked into retrieval. When pdf-inspector's verdict is `image_based` and an OCR request actually came back with nothing usable, the reader refuses the local text layer and the document takes the error state with a message naming the cause. "Retry extraction" now re-reads the pages with OCR for a document whose previous extraction failed or produced unreadable text, the viewer offers that control for poorly extracted documents, and a re-ingestion replaces the previous chunks so retrieval stops answering from the old text.

The gate is the classifier's verdict, not the non-letter ratio, as the issue's measurements argue. Every non-`image_based` verdict, and every deployment with no OCR endpoint configured, keeps today's reader behaviour exactly; `OcrUnavailableError` still propagates for the #633 retry.

## Changes

- `document_readers.py`: `_classify_pdf` runs once and feeds both the local Markdown fast path and `_text_layer_untrustworthy`. The gate sits after the OCR-accept block and before the PyMuPDF fallback, and fires only when the verdict is `image_based` and OCR was actually attempted (`ocr_extract_text_from_pdf` marks `report["ocr_skipped"]` on its two permanent early returns). The refused case marks `report["text_layer_rejected"]`, which the task turns into a specific error message. No confidence threshold on the gate: pdf-inspector 0.2.x emits `image_based` at one fixed confidence, so a threshold never discriminated.
- `force_ocr` (keyword-only, default `False`) threaded through `extract_text_with_markers` into the reader. It means OCR first, not OCR only: the local fast path is skipped up front, and tried again if OCR returns nothing usable or is unavailable, before PyMuPDF.
- `upload_tasks.py` / `document_tasks.py` / `routers/documents.py`: `force_ocr` rides the Celery extraction signature only when `True`. `POST /documents/{uuid}/retry-extraction` sets it when the document is in the error state or `extraction_low_quality`, gains the same `30/minute` limit as upload, returns 409 while an extraction is already in progress, and records `force_ocr` and the previous status in its audit event. Request and response shapes are unchanged.
- `perform_semantic_ingestion` deletes the document's existing chunks before adding the new ones (separate commit).
- `scripts/reocr_old_documents.py` passes `force_ocr=True`; it is the remediation path for documents already stored with a garbled layer.
- Frontend: `RawTextModal` and `DocumentViewer` render the Retry extraction control for a poorly extracted document, not only for one in the error state. The chat notice points at that control again.
- Tests: reader paths (`image_based` with empty, short, and good OCR; OCR skipped because unconfigured; `scanned`, `text_based`, and classification-failure paths unchanged; OCR outage still raises, and falls back to local Markdown only under `force_ocr` on a confidently text-based file; `force_ocr` still honours the blank-page guard and clears a stale partial flag), the task's two error messages, the wire shape, the route's decision, 409, and audit detail through the ASGI app, delete-before-add on ingestion, and a `RawTextModal` test for the new control.

Costs and limits, stated plainly:

- A slide deck or figure-heavy document gets the `image_based` verdict too. With OCR configured it is read by OCR as before; only when OCR ran and returned under 100 characters does it now error instead of storing its captions, the outcome a plain scan already gets.
- The classifier is the gate the issue measured on its two real garbled documents. A garbled layer that pdf-inspector calls `text_based` still takes the local fast path and is flagged only by the existing non-letter-ratio notice (the issue's consequence 2). Three constructed CID-mangled PDFs all classified `text_based` at 1.0, so no fixture exercises the real classifier on this class; the gate tests stub the verdict.
- A forced retry that succeeds stores OCR text with interpolated page markers where the fast path had exact ones. It is only forced for documents whose stored text was already missing or unreadable.
- Chat attachments, workflow fetch, and upload validation read PDFs through the same function and receive an empty string for a refused layer, as they already do for a blank PDF.
- The 409 on an in-progress document is new observable behaviour for API callers; the frontend only offers the control when no extraction is running. It also closes the API-level way to re-dispatch a document whose worker died hard (`processing` stuck true); the reaper already handles the common stuck shapes, and I can add a staleness escape if you want one.
- The route imports `_IN_PROGRESS_TASK_STATUSES` privately and lazily inside the handler, next to the existing lazy `dispatch_upload_tasks` import, to keep Celery out of the router's import graph rather than promote the name for one use.

## Test Plan

- [ ] Shared checks pass (`make ci`) — backend legs run locally: `make backend-static` clean, `make backend-ci` 4810 passed / 161 skipped, coverage 66% against the 50% gate, tier-1 integration passed. Frontend legs (`vitest`, `tsc -b`, `eslint`) run locally in a network-isolated container from the existing `node_modules` (no npm on the host): 823 tests pass, typecheck and lint clean. `frontend-audit` and `frontend-build` are left to CI.
- [ ] Manually tested the affected feature(s) — not done; the two real garbled documents from the issue are not on this machine, and no constructed fixture reproduces the classifier's verdict. Behaviour is pinned by the tests above.
- [ ] Updated deploy/release docs (`README.md`, `DEPLOY.md`, `OPERATIONS.md`, or `RELEASE_CHECKLIST.md`) if the install or release path changed — not applicable; the no-OCR fallback described in `DEPLOY.md` still holds.

## Related Issues

Closes #858


ViaJables:
Reviewed and pushed one commit. Gating on the classifier's verdict rather than the non-letter ratio is the right call and the measurements in #858 make the case; running `_classify_pdf` once and feeding both the fast path and the gate is the tidy shape for it, and the replace-chunks-on-re-ingest fix closes a hole that predates this PR. I verified the `OcrUnavailableError` contract for #633 still holds on every non-forced path.

**What I changed: the gate now needs a positive signal.** `not report["ocr_skipped"]` only distinguishes two of the three ways `ocr_extract_text_from_pdf` returns `""`. The never-configured and undecryptable-key returns set it; the third — attempts exhausted against a *permanent* status (`PERMANENT_STATUS_CODES`: wrong key → 401, oversized deck → 413) — sets nothing, and neither does the reader's `except Exception: ocr_text = ""` fallback. So on a deployment with a bad OCR key, every `image_based` PDF errored with "OCR could not read the pages" where it used to extract through PyMuPDF — a regression on today's behaviour, and a misdiagnosis of the cause. `ocr_extract_text_from_pdf` now marks `report["ocr_completed"]` only when a conversion actually came back, and the gate keys on that. The `ocr_skipped` marker stays as-is; the two tests that model "OCR ran and returned nothing" set `ocr_completed` through a small helper, and there are new tests for the permanent-failure and raising cases.

**The 409 guard locked out the shape this route writes.** `processing=True, task_status="extracting", raw_text=""` matches neither the reaper's `processing: False` nor its `raw_text != ""` clause, so a worker that died after the lock write left the document at "Reading text…" with the only manual re-dispatch answering 409 forever. Extraction start and the retry route both stamp `updated_at` when they take the lock now, and a lock older than 30 minutes yields to a retry. A red-then-green test pins it.

**Two smaller ones.** The `delete_document` before re-ingestion is guarded on non-empty text: the extraction task *returns* `""` on a failed read rather than raising, so the chain still reaches ingestion, and deleting there stripped a document that was searchable a minute ago out of retrieval entirely — worse than old text beside an error state. And `DocumentViewer` resets `lowQuality` with the rest of its per-document state, so the banner from the previous document can't offer force-OCR on one that read fine when the next poll rejects.

## Three things I did not change

**Swallowing `OcrUnavailableError` under `force_ocr` re-stores the extraction the retry was meant to replace.** Your measurements say a constructed CID-mangled PDF classifies `text_based` at 1.0 — which is exactly the branch where the local Markdown fast path accepts. So for the targeted document class, a retry during an OCR outage runs the fast path, stores the same mojibake, marks the document complete, and no deferred retry happens; raising would have hit `TRANSIENT_EXCEPTIONS` and retried with backoff. I can see the argument for a text-based PDF not failing the task over a blip, but for the #858 class it reproduces the loop the PR set out to break. Your call — I'd lean toward raising when `force_ocr` was set *because* the previous extraction was low-quality, and only swallowing when it was a plain re-read.

**The project-KB mirror never refreshes.** `_mirror_into_project_kb` returns early when a `knowledge_base_sources` row already exists for the document, so a document inside a Project keeps answering from the old chunks in "chat with this project" after a successful retry. Pre-existing limitation of the mirror rather than this PR, but it means the delete-before-add only fixes half the retrieval paths. Separate PR.

**The low-quality banner can't render for a PDF.** `setLowQuality` is only called from the `pollStatus` handlers inside the `isDocx` branch, and the banner is inside `if (isDocx)`, so a garbled PDF — the motivating case — takes the `isPdf` path and the viewer offers no Retry control for it; only `RawTextModal`'s new control is reachable. Where that banner belongs in the PDF layout is yours to choose, so I left it.

Merging once CI is green.

PR #900, opened 2026-09-18
fix(extraction): a retry forced by a refused text layer requires OCR

arhyneRWU:
## Summary

Follow-up to #887, the first of the three things left open in its review. A "Retry extraction" that was forced because the document's stored text was low quality could still finish by storing a non-OCR reading of the same pages: with the OCR service unreachable the reader fell back to the local text layer, stored the same mojibake, marked the document complete, and because nothing was raised the extraction task never retried it. The retry route now distinguishes its two reasons. A retry of a document whose text was refused as low quality requires OCR: an outage propagates to the task and takes the #633 backoff, and a conversion that comes back with nothing usable refuses the local layer regardless of the classifier's verdict, so the document takes the error state whose message names the cause. A retry of a document in the error state is a plain re-read with its usual fallback, unless the document's text layer was refused before, which the document now remembers, so the click after a failed OCR-required retry still requires OCR rather than re-storing the garbled layer.

## Changes

- `document_readers.py`: keyword-only `ocr_required: bool = False` on `extract_text_with_markers` and the two PDF helpers beneath it. The outage swallow becomes `if force_ocr and not ocr_required:`, and a generic exception around the OCR call propagates under `ocr_required` instead of becoming an empty result. The garbled-layer gate fires when the verdict is `image_based` *or* `ocr_required`, still only under the positive `ocr_completed` signal from your review commit, and records which half fired so the task's error message does not claim a font diagnosis for a `text_based` document.
- `models/document.py`: `text_layer_rejected: bool = False`. A retry that requires OCR is a fact about the file, and the failure it can produce (an outage that exhausts its retries) clears the low-ratio evidence that set it. Without a durable flag the next click is a plain re-read that re-stores the garbled layer, and the click after that requires OCR again, a loop with no exit. The route sets the flag when it dispatches an OCR-required retry, the task sets it when the reader refuses a layer, and the route reads it back. Default `False`, so no migration.
- `document_tasks.py` / `upload_tasks.py`: `ocr_required` rides `perform_extraction_and_update` and the Celery extraction signature the same way `force_ocr` does, present on the wire only when `True`, so a message enqueued before this deploy is still readable.
- `routers/documents.py`: `retry_extraction` sets `force_ocr` for the error state or a low-quality result, and `ocr_required` for a low-quality result or a remembered refusal; both only for PDFs, since the task ignores them for other formats and the audit record should not claim otherwise. The audit event records both. Request and response shapes are unchanged.
- Tests: the reader under `ocr_required` raises on an outage, refuses an empty and a short conversion on a `text_based` verdict, refuses without any classification, accepts a good conversion, is inert when OCR is unconfigured, and still honours the blank-page guard; `force_ocr` alone keeps its fallback on the same shapes; the task forwards the flag; the wire carries it only when set; the route dispatches and audits both flags for each reason.

Deliberate limits:

- A permanent OCR failure (bad key, oversized file) under `ocr_required` keeps the PyMuPDF fallback, for the reason your review commit gave: erroring there would misdiagnose a misconfiguration as unreadable pages. The document re-stores its text and stays flagged low-quality.
- A deployment with no OCR endpoint configured is unchanged; `ocr_required` has no effect there.
- `scripts/reocr_old_documents.py` keeps `force_ocr` alone. It selects documents by age, not quality, so it is a bulk plain re-read.
- No `**kwargs` shim on the task for an old worker receiving the new key during a rolling deploy, matching how `force_ocr` was introduced. Truthy-only sending narrows the window to OCR-required retries during the restart; the chain's `link_error` moves such a document to the error state within seconds, and the next click self-heals.
- `ocr_required` implies OCR-first in the reader, so the pair the route never constructs (`ocr_required` without `force_ocr`) cannot silently take the local fast path.
- A genuinely sparse PDF (a cover sheet, say) retried under `ocr_required` errors when OCR finds under 100 characters, where before it would have stored its few words. It is only retried that way if its stored text was already refused as low quality.

## Test Plan

- [ ] Shared checks pass (`make ci`) — backend legs run locally: `make backend-static` clean, full backend suite 4909 passed / 161 skipped. Frontend legs untouched by this change and left to CI.
- [ ] Manually tested the affected feature(s) — not done; the outage path is pinned by the reader and task tests, and the real garbled documents from #858 are not on this machine.
- [ ] Updated deploy/release docs (`README.md`, `DEPLOY.md`, `OPERATIONS.md`, or `RELEASE_CHECKLIST.md`) if the install or release path changed — not applicable.

## Related Issues

Follow-up to #887 (review, "Swallowing `OcrUnavailableError` under `force_ocr`"). Relates to #858 and #633.


ViaJables:
Reviewed — core logic holds (the two-flag split, both gates, the persisted refusal, the truthy-only wire encoding). Pushed 059492b4 on top with three edge-case fixes rather than bounce them back as follow-ups:

1. **`text_layer_rejected` was permanent.** The route sets it from the non-letter ratio before OCR has confirmed anything, so a ratio false positive (a short page of checkbox glyphs) pinned the document to OCR-only re-reads for life — and with OCR down, a retry on a document whose local reading was fine now sat through the backoff and landed in error where it used to be stored immediately. The extraction task now clears the flag when it stores a successful reading; the next retry starts from status and ratio like any other. Model comment updated to match.
2. **The `except Exception: if ocr_required: raise` didn't reach the retry.** The task's catch-all records anything that isn't `OcrUnavailableError` as "Text extraction failed" and returns before `autoretry_for` sees it, so the backoff the comment promised never ran. Now re-raised as `OcrUnavailableError` (with `__cause__`), which the existing handler retries and, out of retries, reports as an OCR-reach failure.
3. **The `ocr_required` message foreclosed the retry.** The reader pops `report["partial"]` before the task sees it, so "found nothing usable" can be one page of three on a Docling `partial_success`. The message now says "Retry extraction in case the OCR service was degraded, or re-upload…".

Each change has a test that goes red without it. 225 pass across the four touched test files.

PR #901, opened 2026-09-18
fix(ingestion): a project's KB chunks are refreshed when the document is re-extracted

arhyneRWU:
## Summary

Follow-up to #887, the second of the three things left open in its review: the project knowledge-base mirror never refreshed. `_mirror_into_project_kb` returned as soon as a `knowledge_base_sources` row existed for the document, so a document inside a Project kept answering "chat with this project" from its old chunks after a successful retry, even though #887 made the document's own collection replace them. The early return is now a currency check: a row whose `content_hash` matches the new text, is `ready`, and has chunks is left alone; anything else has its chunks deleted and re-added, its row updated, and the knowledge base's counters recomputed from the rows. A refresh that fails after the delete marks the row as errored rather than leaving it claiming to be indexed.

## Changes

- `document_tasks.py` `_mirror_into_project_kb`: the existence check becomes a currency check using the existing `kb_source_currency` helpers. Refresh path: `delete_kb_source` (project KBs key chunks by document uuid) → `add_to_kb` → `update_one` on the row with the new chunk count, `ready`, and the ingestion stamp → `_recalculate_kb`, which recomputes counts from rows and cannot double-count. The insert path is unchanged except that it now writes the ingestion stamp too, so a later move of an unchanged document is a no-op. Rows written before this change have no fingerprint, so they refresh once and then settle; no migration.
- Failure path: `delete_kb_source` now reports whether the delete ran, because it swallows its own Chroma errors and `add_to_kb` adds under deterministic ids; a failed delete followed by a successful add would have left the old chunk text under a row stamped with the new fingerprint, which the gate would then certify forever. A refresh whose delete or add fails marks the row `error` with no chunks, recomputes the counters, and sends the existing mirror failure notification with a body that says which of three things happened: the previous copy could not be removed, so the project may still answer from the older text; it was removed and could not be replaced; or, on a first indexing, the document could not be added. Each names Retry extraction as the remedy.
- Tests: `delete_kb_source` reports success and failure; a failed delete stops the refresh before the add and errors the row; unchanged text is not re-embedded; changed text deletes before it adds and updates the row; a legacy row without a fingerprint refreshes once; an errored row refreshes even when the text matches; a row claiming zero chunks refreshes; a move with unchanged text costs nothing; the insert branch stamps the fingerprint; a failing add after the delete leaves the row errored and still raises; the ingestion task reaches the refresh branch end to end.

Deliberate limits:

- A refresh that fails is not retried by the ingestion task. The mirror's existing contract catches, notifies, and returns, and I kept it. The consequence is that project chat goes from stale answers to none for that document until the next Retry extraction, the same shape #887 accepted for the document's own collection. The error row and the notification are the recovery path.
- When it is the delete that fails, the old chunks are still in Chroma while the row says errored with no chunks, so retrieval keeps answering from the old text until the next Retry. The notification says so; the row under-counts for that case.
- Nothing locks the mirror against a concurrent `kb_reingest` of the same project knowledge base. Both delete then add under deterministic ids keyed by document uuid, so the outcome is a truncation window rather than duplicate chunks, and that window predates this change.

- Explicit knowledge bases that also contain the document are not refreshed by this change. `_resume_pending_kb_sources` only re-queues pending rows, so a `ready` source in an explicit KB keeps its old chunks after a retry. Refreshing those fans out one embed job per KB per retry, so it is a separate decision.
- Only the refresh path uses `_recalculate_kb`; the insert path keeps its `$inc`, so counters that have already drifted are repaired only for documents that get refreshed.
- `delete_kb_source` swallows its own Chroma errors, so the guarded pair is effectively `add_to_kb`; the guard is about the state the row is left in.
- The currency check fingerprints the text only, as the shared helper does for explicit knowledge bases. A re-read that yields identical text with different page markers is a no-op for the project KB while the document's own collection refreshes.

Noticed but not touched: the knowledge-base source-add route and `knowledge_base_tasks.py`'s source ingestion lack the implicit-KB branch that `kb_reingest` has, so a project knowledge base reached through the KB API would key chunks by source uuid rather than document uuid. No endpoint exposes a project's KB uuid today, so it is unreachable, but it is the shape that would put two copies of one document into retrieval.

## Test Plan

- [ ] Shared checks pass (`make ci`) — backend legs run locally: `make backend-static` clean, `make backend-ci` 4904 passed / 161 skipped, coverage 66.61% against the 50% gate, tier-1 integration passed. Frontend untouched, left to CI.
- [ ] Manually tested the affected feature(s) — not done; pinned by the pymongo-mock tests in the style of the surrounding mirror tests.
- [ ] Updated deploy/release docs (`README.md`, `DEPLOY.md`, `OPERATIONS.md`, or `RELEASE_CHECKLIST.md`) if the install or release path changed — not applicable.

## Related Issues

Follow-up to #887 (review, "The project-KB mirror never refreshes"). Relates to #858.


ViaJables:
Reviewed — the currency gate → delete → add → stamp → recount sequence is sound, and the hash is the same `raw_text` extraction wrote, so unchanged-document moves genuinely no-op. Pushed 1ed8fea7 on top with two fixes:

1. **A successful refresh could be errored by its bookkeeping.** The guard spanned the row stamp and `_recalculate_kb` as well as the delete/add pair, so a transient Mongo error *after* `add_to_kb` succeeded stamped the row `error, chunk_count: 0` and belled "its previously indexed copy was removed and could not be replaced" — when the new chunks were in Chroma and current. The guard now covers only delete/add; the stamp and recount log on failure and leave the row alone (a failed stamp keeps the old fingerprint, so the next pass refreshes again). `test_a_failure_after_the_add_is_still_a_refresh_failure` was pinning the old behaviour and is replaced by two tests for the new one.
2. **`_remove_from_project_kb` decremented `sources_ready` blindly.** Project-KB rows can now be `status: "error"`, and moving such a document out of the project left `sources_ready` one low and `sources_failed` one high until some other document happened to trigger a recount. It now calls `_recalculate_kb` after the `delete_one`, same as the refresh path.

Both tests go red without the fix; 70 pass in `test_document_tasks.py`.

